PR #2, opened 2016-12-09
add tidal app

zeke:
A followup to https://github.com/electron/electron.atom.io/pull/550


zeke:
Gonna go ahead and ship this. 🎵

PR #8, opened 2016-12-09
Add VK Messenger app

gch1p:
I am the developer of **VK Desktop Messenger**, which we have launched a few days ago. **[VK.com](https://vk.com)** is the largest European social network based in Russia.


zeke:
Thanks!

PR #20, opened 2016-12-09
add NOTR

zeke:
A followup to https://github.com/electron/electron.atom.io/pull/593


dhaval17:
@zeke Thank you

zeke:
> Anything else required from my end ?

Nope, all set. This will land on the website once we ship https://github.com/electron/electron.atom.io/pull/616

PR #23, opened 2016-12-12
update pull

foozzi:


zeke:
@foozzi this PR only changes whitespace. Was that your intention?

foozzi:
@zeke yes, I was not correctly named icon

PR #26, opened 2016-12-15
add fangyuanjian

yanue:


zeke:
Thanks for the submission, @yanue. We would be happy to include this app on the website, but first I have a request.

Because this is the first non-English app that's been submitted, I would like to suggest adding a `locales` property to the YML file to denote that this app is only available in Chinese. The `electron/electron` docs have translations in both `zh-CN` and `zh-TW`. Which of those make sense for "fangyuanjian"?

I'm thinking maybe something like this:

```yml
name: fangyuanjian
description: '办公即时通讯,自由办公,组织架构'
website: 'http://bzsns.cn/'
keywords:
  - "即时通讯"
  - "组织架构"
  - "场地服务"
  - "自由办公"
  - "组织架构"
locales:
  - "zh-CN"
```

cc @zcbenz and @Stone-Jay, @ArisLittle, and @sqrthree who have all helped out translating Electron's docs to Chinese.

Stone-Jay:
```
locales:
  - "zh-CN"
```

ArisLittle:
For example, the string in description key:
'办公即时通讯,自由办公,组织架构' - This is `zh-CN`
'辦公即時通訊,自由辦公,組織架構' - This is `zh-TW`
so the `zh-CN` makes sense

yanue:
@zeke i have added a locales line, additionally i also can translate it into English.

zeke:
> additionally i also can translate it into English.

That would be great! How about something like this:

```yml
description: "办公即时通讯,自由办公,组织架构 (Office instant messaging)"
```

.. or whatever you think is a good translation.

Thanks for being patient as we figure out how to handle multiple languages. 👍

yanue:
hi @zeke , i have changed all into english.

PR #33, opened 2016-12-21
add cloudtag

fedegratti:
Hello, I did a desktop app using electron to share files easily. I want to show it in your official website. Thank you very much.


zeke:
Thanks!

Your app will be live on [electron.atom.io/apps](http://electron.atom.io/apps/) within a few minutes.

PR #46, opened 2016-12-31
Add files via upload

MedZed:
added my electron app (youtube-mp3) .. a youtube to mp3 converter


zeke:
A few issues:

screen shot 2016-12-31 at 6 46 06 am

MedZed:
The icon issue 3) is it square .. its 512 x 512 should I change the transparent background or lower the dimensions ?

PR #53, opened 2017-01-06
Add Shift

rollerbrick:


zeke:
Nice icon!

The description is a bit long. How about "Switch between Gmail, Calendar and Drive accounts with ease"? At the very least, "Shift is the " should be removed.

PR #55, opened 2017-01-07
Add MongoDB Compass

imlucas:
https://mongodb.com/compass


zeke:
Nice!

I would recommend breaking down the keywords into `mongodb`, `GUI`, etc...

imlucas:
@zeke Happily. See latest.

zeke:
🍃

PR #58, opened 2017-01-11
Add Shapespark

wrr:


zeke:
Thanks for submitting! 

Instead of leaving the `keywords` blank, please add a few like `WebGL` and `3D`, or just remove the property entirely, as it is not required.

wrr:
Thank you for the review, please take a look again

zeke:
👍

PR #59, opened 2017-01-11
Resize app icons at install time

zeke:
## Problem

There are 6.5 MB worth of icons on [electron.atom.io/apps/](http://electron.atom.io/apps)

## Solution

This PR adds a script that resizes all the app icons to various smaller dimensions: 32, 64, 128.

To keep the repo small and free of auto-generated content, this new resize script is run in an npm postinstall hook:

```
$ npm install electron/electron-apps

$ tree node_modules/electron-apps/apps     
node_modules/electron-apps/apps
├── 1clipboard
│   ├── 1clipboard-icon-128.png
│   ├── 1clipboard-icon-32.png
│   ├── 1clipboard-icon-64.png
│   ├── 1clipboard-icon.png
│   └── 1clipboard.yml
...
```

We can shave off 5 MB by using the 64x64 images on the apps page:

```
❯ du -hc apps/**/*icon.png | grep total
6.5M	total

❯ du -hc apps/**/*64.png | grep total
1.5M	total
```

Even if we use 128px icons we can shave off 3.4 MB:

```
du -hc apps/**/*128.png | grep total
3.1M	total
```

## Notes

I initially tried this with [jimp](https://github.com/oliver-moran/jimp) but it couldn't handle all the images. Now using [sharp](http://sharp.dimens.io/en/stable/api-resize/) which introduces a native dependency, but it's fast and seems to work well for all the existing 250-ish icons.

cc @electron/maintainers

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 yarn.lock
 index.json
 node_modules
+*-32.png
+*-64.png
+*-128.png
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,11 @@
+addons:
+  apt:
+    sources:
+    - ubuntu-toolchain-r-test
+    packages:
+    - g++-4.8
+env:
+  CXX=g++-4.8
 language: node_js
 node_js:
   - "4"
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "A collection of apps built on Electron",
   "main": "index.json",
   "scripts": {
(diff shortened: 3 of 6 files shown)


zeke:
@electron-bot handled this update like a champ: https://github.com/electron/electron.atom.io/commit/2d3fd1265d8e374be97725f4237488c22f7add7a

Next step: actually use the icons on the apps page to cut down the download size.

PR #60, opened 2017-01-12
Add Papyrus

morkro:


zeke:
Thank you very much for your submission for inclusion in the Electron app list. We're uncomfortable with including your application in the list because of the icon's similarity to the [Dropbox logo](https://www.dropbox.com/branding). Because of this, we've decided to not include your application in our list.

If you're willing to change your application icon, please open a new pull request and we will happily include your app.

Thanks very much for your submission and your use of Electron!

morkro:
Thanks for the info, I will change the logo :)
How does that work with [Caprine](http://electron.atom.io/apps/?q=caprine) then? It's also using the official Facebook Messenger icon and is not an official client.

zeke:
I was unaware that Caprine's logo is using the FB messenger icon.

Thoughts, @sindresorhus? Maybe change the icon to a goat?

🐐

zeke:
Merging and setting `disabled: true`, in case @morkro wants to update the icon at some point.

PR #63, opened 2017-01-13
disable apps using company logos without permission

zeke:
I have consulted with our legal team, and they have advised us to disallow apps that are using the names of other companies or icons that we find too similar to the logos of other companies without verifying their permission to do so.

That said, we would love to continue featuring all of these apps on our site. If you are an author of one of these apps and would like to see it on the site again, please open a PR that updates the icon file and removes the `disabled: true` line from the app's yml file.

Apologies for the inconvenience, and thanks for your understanding.

cc @terkelg @sindresorhus @MedZed @morkro

### DIFF
--- a/apps/caprine/caprine.yml
+++ b/apps/caprine/caprine.yml
@@ -8,3 +8,4 @@ keywords:
     - social
     - facebook
 license: MIT
+disabled: true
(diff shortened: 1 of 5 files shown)


zeke:
Slack HQ reached out to us and asked that "Slack Catchup" be removed. As stated above, please feel free to make the necessary changes (in this case changing the name of the app) and we will happily re-enable your app on the site.

cc @sivaramsi @paulcbetts

terkelg:
No problem, thanks for the heads up @zeke. I'll update the Ramme icon as soon as possible. 👍

PR #69, opened 2017-01-17
Add illyriad

benaadams:
Steam game using electron and [Greenworks](https://github.com/greenheartgames/greenworks)


zeke:
Thanks!

I don't know much about the Steam platform. Is there any special work required to get Electron apps to run on it, or to be accepted in to the Steam store?

benaadams:
Its a little involved...

To be accepted you first have to go through Steam Greenlight where you propose [the idea of your game](https://steamcommunity.com/sharedfiles/filedetails/?id=323081123) for the platform and the Steam using public vote on whether they feel they would play the game.

After that you need to properly integrate with the platform, accounts, etc; and Value do extensive Q&A to ensure the game works and that you have integrated properly before they will let you release.

Luckily https://github.com/greenheartgames/greenworks provides a node wrapper for the native dlls for Electron and NW and provides instructions on how to integrate on the readme page.

However, there are some common issues covered in the [Troubleshooting](https://github.com/greenheartgames/greenworks/wiki/Troubleshooting) section. Most notably is having to have a canvas drawing every frame and running the gpu inprocess, so the Steam overlay works.

Once you've done those bits though it works very well and preforms well.

zeke:
Thanks for the info.

I looks like this game is only available on Windows:

screen shot 2017-01-17 at 10 05 53 am

Can Steam games run on other platforms?

benaadams:
Yes, we are looking to bring out mac and linux electron-based versions after we've ironed out any launch bugs.

zeke:
👍

PR #70, opened 2017-01-17
Add ReqView

liborbus:


zeke:
Thanks, @liborbus. The term "requirements management" is new to me, and probably others as well. I would suggest updating the description to use more familiar terms.

That said, this is just a suggestion. If you want me to merge this as-is, I will. 👍

liborbus:
@zeke, you are right. This keyword doesn't make much sense for categorization of electron apps. I removed it.

PR #71, opened 2017-01-19
Update destroyer.yml

mashaal:


zeke:
FYI the new search [includes keywords](https://github.com/electron/electron.atom.io/blob/6d4a6d6785e4d049ab439ada05f6e761342437a8/_pages/apps.html#L27-L29), even though they aren't visible on the page. I would suggest sticking to the old title without '/Music', as a search for 'music' will still find this app because of the newly added keyword.

mashaal:
thanks! the new search looks great by the way 😇

zeke:
Glad you like it. We'll keep iterating. :)

PR #73, opened 2017-01-19
Adding Zector to Electron apps directory

jay3sh:
Zector is suite of Vector graphics tools designed for graphics
artists. Each application in this suite is built using Electron.
The applications are proprietory, but are available for download
for the purpose of proving that they are built on top of Electron.

You can download from this URL: http://zector.io/download


zeke:
Thanks!

PR #75, opened 2017-01-20
Added Cerebro app

KELiON:
Yet another cross-platform extendable open-source launcher. It is built with js, Electron and React. Cerebro is focused on extendability, speed, good UI and UX. In most cases, you don't need to launch another application – you can see everything in Cerebro, i.e. google maps, films on IMDB, results of shell scripts or gifs


zeke:
👍

PR #76, opened 2017-01-20
Add opale-messenger app

DavidMellul:
Hi ! 

I'd like my project to be added to this list. 
The reasons why it should be included : https://github.com/DavidMellul/Opale-Messenger
In summary, it's a wrapper for Facebook Messenger, but it comes with the android-like bubble.

Thank you in advance.
Cheers.


zeke:
Thanks for submitting. What is an "android-like bubble"?

DavidMellul:
Hey !
By android-like bubble I mean, when you're on Messenger on your Android, you have a bubble (maybe it's not like that it's called ? ).
Here is an image of what I'm talking about : 
image

Thank you for your consideration.

PR #77, opened 2017-01-23
Add Coypu

bkzl:
The idea behind Coypu is to make a week planner with the text-editor-like interface to input data. It means you can f.e. add new tasks using enter, jump between todos with arrow keys, use backspace to remove existing entries or complete tasks with the tab key.

Free, actively developed and works offline. Built with Vue.js, Electron and PouchDB.

👉 https://coypu.co


zeke:
Looks nice!

PR #84, opened 2017-02-02
Rename "Sync" to "Wantedly Chat"

yamadayuki:
Hi, we asked to list our app to electron.atom.io before. Now Sync is renamed Wantedly Chat.

Cc. @luvtechno
ref. https://github.com/electron/electron.atom.io/issues/136


yamadayuki:
Thanks!

PR #91, opened 2017-02-10
Add appear.in

vitorgalvao:
As per the rule:

> Don't use another company's trademarks (icon, logo or name) without supplying evidence of prior permission

https://github.com/vitorgalvao/appear.in/pull/1 proves I have their permission.


zeke:
> We're fine with you using our logo and name for this, as long as it's made clear that the application is unofficial, now and in the future.

Thanks for clarifying 👍 

Would you ming adding a few more keywords to increase discoverability on the website? Maybe `p2p`, `video`, `chat`, `collaboration`, `conference`, etc

vitorgalvao:
Added a few more keywords.

zeke:
🤘

PR #95, opened 2017-02-12
Add Lepton as featured app on Electron website

hackjutsu:
Here is our app **Lepton**.

It is a GitHub Gist client built on Electron. We provide free macOS/Windows/Linux downloads on our [release](https://github.com/hackjutsu/Lepton/releases) page.

Please let me know if there is something missing in this PR.

Thanks,

-Cosmo


apps/lepton/lepton.yml
```diff
@@ -0,0 +1,14 @@
+name: Lepton
+description: 'Open-source GitHub Gist client built for macOS, Windows and Linux. '
```
zeke: There's an extra space at the end of the description. Could probably live without the "for macOs... " part too.

hackjutsu: Thanks for feedback. I'll remove the extra space and the "for macOS part".

hackjutsu:
@zeke Haha, that's interesting~ 

That's my English name I picked up around 2008 before I moved to US. 
Are you living in SB? Me too!

zeke:
> Are you living in SB?

I was! But I moved yesterday. Moved down to Moorpark / Thousand Oaks with my family.

Next time I'm in SB we should have coffee. ☕️

hackjutsu:
@zeke Sure thing!

PR #107, opened 2017-02-20
remove SymbSearch

leun4m:
SymbSearch **is not** deprecated, but not longer built in electron, since it uses to much ressources for a small tool like this. So, to be fair, it cannot appear anymore under apps listed as electron-apps.

However, thank you for listing it up :smiley:


zeke:
Sorry to hear that Electron wasn't working out for you.

What kinds of issues were you running into specifically?

How are you building the app now?

What was the last version of Electron you were using?

It might be too late for you to switch back, but just so you know: the Chrome team has been very focused on reducing Chromium's resource footprint lately. Since Chrome 53 (Electron 1.4), battery life has been extended by as much as 33%: https://blog.google/products/chrome/chrome-faster-and-more-battery-friendly/

leun4m:
Well the problem is mostly the big disk space:
As executable program made by the `electron-packager` it takes `100MB`. By using Visual C# now I can get the whole functions with only `250KB`.

Of course I loose the platform indepence now I would like to have, but if the price is so high, I  will not pay it. — If you or somebody else might have a better idea, I indeed would like to know it.

One of my priorities is indeed: SymbSearch should be a a slim tool.

---

I used:
- electron: **1.3.5**
- electron-packager: **8.1.0**

zeke:
Hey @malept, any tips for slimming a Windows build?

malept:
@zeke nope. Here are all of the file in `electron-v1.3.13-win32-ia32.zip` that are more than 1 MB:

```
 60M Dec  6 14:46 electron.exe
 13M Dec  6 14:41 node.dll
9.7M Dec  6 14:36 icudtl.dat
9.4M Dec  6 14:36 content_shell.pak
3.4M Dec  6 14:37 d3dcompiler_47.dll
2.2M Dec  6 14:36 libGLESv2.dll
1.9M Dec  6 14:36 ffmpeg.dll
1.4M Dec  6 14:34 LICENSES.chromium.html
```

This ends up being ~101MB. So this is more of an Electron problem in general, rather than a problem with how Electron Packager bundles up the app.

zeke:
> this is more of an Electron problem in general, rather than a problem with how Electron Packager bundles up the app.

For sure. I just pinged you because I knew you'd know the scoop. 👍

PR #108, opened 2017-02-20
Add Soundkeys

MSandeep96:


zeke:
How about just "SoundCloud wrapped in a desktop player" as the title?

MSandeep96:
That description drops the purpose of the app.

MSandeep96:
Choos

PR #114, opened 2017-03-01
add script for cleaning leftover build artifacts

zeke:
`npm run build` wasn't working on machine after some apps were removed. This adds a cleanup step to the build process to ensure removed apps are fully removed.


zeke:
[Travis is down](https://twitter.com/traviscistatus/status/837055200204484608). #yolo

PR #115, opened 2017-03-01
fix broken reference to slug

zeke:
This typo was breaking the build process, but it wasn't picked up by CI.

One of the perils of only using CI to test [human-generated content](https://github.com/electron/electron-apps/blob/aacf8bf73a68a99d74b8b0c8189c9998f7b98257/package.json#L12-L13).


zeke:
[Travis is down](https://twitter.com/traviscistatus/status/837055200204484608). #yolo

PR #124, opened 2017-03-08
Add short.cm app

gugu:


zeke:
screen shot 2017-03-08 at 4 52 30 pm

The `website` value should be a URL to webpage, not an installer file.

We may eventually add support for specifying URLS of installers.

apps/shortcm/shortcm.yml
```diff
@@ -0,0 +1,3 @@
+name: Short.cm
+description: 'An app for making short links and screenshots on your custom domain'
+website: 'https://s3.amazonaws.com/shortcm-downloads/Short.cm.dmg'
```
db2k: How about adding keywords to make it easier to search for?

```yml
keywords:
    - links
    - linkshortener
    - screenshot
    - domain
```

for example

gugu:
Thanks!

PR #127, opened 2017-03-10
Adding NeutrinoMetrics to app list

mengfoobar:


zeke:
This looks really interesting. Looking forward to giving it a try.

mengfoobar:
@zeke  thanks :D its still in beta, but it'd be great if you have any suggestions/feedback!

PR #129, opened 2017-03-10
Change the app name Boost to Boostnote

asmsuechan:
Hi, developers! Our team has changed the app name Boost to Boostnote. And also changed its logo. So I'd like to change the content in https://electron.atom.io/apps.


zeke:
Looks good. Thanks.

PR #138, opened 2017-03-17
✨ Added Playcode to apps list

Meadowcottage:
Adding @playcode to apps list 😄 🎉


zeke:
This is awesome!

I would say these keywords should be removed: 

- playcode
- desktop-client

And `online-editor` should just be `editor`

Meadowcottage:
Gone ahead and updated the keywords for you.
Should be good to go! 😄

zeke:
You're fast!

Meadowcottage:
Meh. Quick 2 second fix and I'm already at my PC so it's easy 😋

PR #140, opened 2017-03-31
Add Surf to apps list

alchaplinsky:
[Surf](https://surfapp.io/) - Organize your ideas and inspirations with ease


zeke:
Thanks!

PR #143, opened 2017-04-03
Added WMail

Thomas101:


zeke:
nice icon. 👍

MichaelTunnell:
wait what now?

Why would you suggest WMail be added to the Electron App list on April 3rd when you ended WMail on March 30th? [Source](https://blog.wavebox.io/introducing-wavebox/)

If anything, Wavebox should be submitted.

vitorgalvao:
Agreed with @MichaelTunnell. I just saw WMail on the web page and came here to remove it. I’m surprised to find its developer submitting it when the app was discontinued for another.

PR #147, opened 2017-04-11
Add System Designer

ecarriou:
System Designer is an IDE for designing JavaScript applications driven by the model. 

More informations:

* [website](https://designfirst.io/systemdesigner/)
* [source](https://github.com/design-first/system-designer)
* [electron source](https://github.com/design-first/system-designer-electron)
* System Designer has been published on both [Mac App Store](https://itunes.apple.com/us/app/system-designer/id1102494854?l=fr&ls=1&mt=12) and [Windows Store](https://www.microsoft.com/store/apps/9p05jrx99h87)


zeke:
Thanks. Can you explain what "driven by the model" means?

ecarriou:
In case of System Designer, it means that you define a model in a declarative manner in JSON, then:

* JavaScript classes are generated based on this model,
* components instantiated by these classes are immutable, you can not alter them and
* signatures / returned value of invoked methods are checked on runtime based on the model.

This approach is inspired by what we can find in [MDA](https://en.wikipedia.org/wiki/Model-driven_architecture).

PR #149, opened 2017-04-13
add matrix writer

AndrewLang:
Matrix Writer is a blog editor developed base on electron.


AndrewLang:
@fonghuiyue it seems my changes were not committed to the master.

PR #150, opened 2017-04-16
Revert "Added WMail"

vitorgalvao:
Reverts electron/electron-apps#143.

As stated in #143, WMail no longer exists. It makes no sense to have a dead app on the website.


zeke:
Rather than removing this, how about updating it with the new name, repo, and logo?

cc @Thomas101

zeke:
Just found https://github.com/electron/electron-apps/pull/151. Sorry for the confusion, and thanks for fixing this.

PR #157, opened 2017-04-25
Add Alchemy

mfix22:


zeke:
This looks like a nice app. What operating system(s) does it support?

One little request: Can you remove "Alchemy is" from the description?

mfix22:
hey @zeke!
Currently we have only tested it on macOS. We have an issue open in our repo for Linux and Windows.

Absolutely! I was wondering the same thing. On it.

zeke:
👍

PR #158, opened 2017-04-26
Add Headset

danielravina:


db2k:
How about linking this page instead of the git repo?
http://headsetapp.co/

danielravina:
Hi, I put the page in the `website` key. Is there anything else I need to do?

zeke:
Thanks!

PR #160, opened 2017-04-28
Matrix writer

AndrewLang:
Add Matrix Writer


zeke:
Already landed in https://github.com/electron/electron-apps/pull/154

PR #165, opened 2017-05-05
Add Figma

bfarooqi:
The [Figma Desktop App](https://www.figma.com/downloads) is built using Electron.


zeke:
Of course! Surprised this wasn't in here already.

cc @poiru

PR #166, opened 2017-05-07
Add todometer

cassidoo:
Adding my app todometer, built with Electron. :)


zeke:
This looks great! Just one little issue: the website needs to be a fully-qualified URL.

screen shot 2017-05-06 at 10 16 07 pm

zeke:
✅

PR #172, opened 2017-05-16
Github Desktop

zeke:
https://desktop.github.com/

https://github.com/desktop/desktop

https://githubengineering.com/how-four-native-developers-wrote-an-electron-app/

cc @joshaber @shiftkey @niik @donokuda


zeke:
Looks like Preview.app did something weird when I extracted that PNG from the ICNS.

@donokuda can you hook me up with a better PNG?

niik:
> @donokuda can you hook me up with a better PNG?

I gotchu

PR #176, opened 2017-05-19
Adding Circuit app

alper-teke:
Adding Circuit Desktop app which uses electron


zeke:
Thanks. Would you mind adding a few keywords to make it clearer what the app is for?

alper-teke:
Yes sure..

zeke:
👌

PR #178, opened 2017-05-22
add mstream

IrosTheBeggar:
mStream is music streaming server that uses Electron to add GUI management tools


zeke:
Nice! Want to add some keywords like `music`, `audio`, and `player` to make your app easier for folks to find?

IrosTheBeggar:
I added some keywords and a license

zeke:
👍 Thanks

PR #180, opened 2017-05-27
Adding "EasyTongue" to Electron apps

FlorianFe:
Adding the translation app "EasyTongue" to the list of example Electron apps


zeke:
Thanks!

PR #188, opened 2017-06-05
added colol app

Slllava:
colol app


zeke:
Nice!

zeke:
You might want to open another PR that adds some keywords like 'design' and 'tool' to make this easier for folks to discover.

PR #189, opened 2017-06-06
Add streetviewdownload360

ThomasOrlita:


zeke:
Is this an Electron app? I couldn't find a download link at that URL.

ThomasOrlita:
Yes, it is. You can download it when you click the “Download 360°” button (in the top of Street View). Here's a link to the app: https://streetviewdownload.eu/download

PR #190, opened 2017-06-07
added autoEdit app to list of electron apps

pietrop:
added [autoEdit.io](http://autoedit.io) app to the list of electron apps.

I think it be a great show case for the electron community of an example of text based video editing that makes video production faster, easier, and more accessible.


zeke:
That icon is going to look small because it has so much empty space around it. I would suggest scaling it up to fit the bounding box.

Also it looks like the tests are failing.

> human-submitted app data autoedit icon exists as autoedit-icon.png:
> autoedit-icon.png not found

pietrop:
Thanks, ok I've modified the image name and size, let me know if this is good to go or if there's anything else I can do.

zeke:
Sorry, the icon has to be square and between 256px and 1024px in dimension.

pietrop:
Sorry, for the back and forth, is this better?
It is now the same size as the one for `atom` in that folder.

zeke:
👍

pietrop:
awesome thanks!

pietrop:
I don't see it in the main site under [https://electron.atom.io/apps/?q=video](https://electron.atom.io/apps/?q=video) , is there anything else I need to do?

zeke:
@pietrop it looks like there is a hiccup in our deployment. I'll let you know when it's live.

zeke:
@pietrop your app is now on the site!

FYI, here's the little bug that was blocking it: https://github.com/electron/electron-apps/pull/193

pietrop:
Awesome @zeke ! thanks!

PR #197, opened 2017-06-13
Add CashNotify

julienma:
Add [CashNotify](https://cashnotify.com/) to Electron apps, as suggested by @zeke on https://github.com/sindresorhus/awesome-electron/pull/459#issuecomment-308165828.


zeke:
😎

PR #199, opened 2017-06-15
Add Hoster

Houfeng:


zeke:
Thanks, @Houfeng. It looks like this app is not in English. Can you add a `locales` array to your YML file? I think this is Chinese simplified.. right? If so, it should look like this:

```yml
locales:
  - zh-cn
```

zeke:
You might want to add some other keywords too like `development`, `hostnames`, `ip address`, `dns`, etc.

Houfeng:
@zeke 
Locales has been added, and the keywords have been updated

thx

PR #200, opened 2017-06-16
Sloth is a clipboard manager that understand what you copy

moomou:
Sloth is available via https://ohsloth.com and uses machine learning as well as program context to help you find snippets faster.

Here is a screenshot of the electron as a dependency: 
screen shot 2017-06-15 at 11 58 56 pm


zeke:
This looks really cool. You might want to fill in the background of your icon so it's not transparent. You should probably also make the icon fill more of the dimensions so it's doesn't look smaller than other icons.

moomou:
Thanks for the suggestions! Updated.

PR #201, opened 2017-06-19
Adding the App "Elements"

FlorianFe:
"Elements" is a desktop App which displays the periodic table.


zeke:
This is beautiful! 😍

PR #205, opened 2017-06-23
Add Perlotto app

can3p:
[Perlotto](https://github.com/can3p/perlotto) app is a wrapper around google music web service that turns it in to desktop app.

Additional features:
* Multimedia keys support
* Last.FM support


zeke:
Cool. Kinda like https://www.googleplaymusicdesktopplayer.com/

PR #206, opened 2017-06-24
jqi app

nire0510:
The almighty jq processor - wrapped in a graphical UI


zeke:
Wow this is a cool idea. I would suggest adding some keywords like json, developer, tool, editor, etc to make your app easier for people to find.

nire0510:
@zeke thank you very much!
You're absolutely right. Just added few keywords.
Thanks!

PR #210, opened 2017-06-28
cliptext added

aaabhilash97:
#### ClipText

ClipText is a simple clipboard manager for mac.
###### Features
 - Save history of your last copied texts
 - Global hotkey for popingup tray context menu with history.
 - Global shortcut for launching Clipboard history is Cmd+Alt+h

 Download link : https://github.com/aaabhilash97/cliptext/blob/master/dist/cliptext-1.0.6-mac.zip


zeke:
Hi @aaabhilash97. It looks like there's something wrong with your icon file:

screen shot 2017-06-28 at 1 14 00 pm

zeke:
Nice. Can you also change the description? How about `A simple clipboard manager for macOS` -- no need to mention Electron, or the name of the app, and OS X is officially called macOS now.

zeke:
👌

PR #217, opened 2017-07-02
Adding Themer to apps list

mjswensen:
Hello! I'd like to add this app to the app registry at atom.electron.io/apps, since it is built on Electron. ❤️

An introduction to the app can be found here: https://themer.mjswensen.com


zeke:
🌈

PR #219, opened 2017-07-07
Add Oversetter app

4gray:
Oversetter is a simple translation app, which allows you to translate content directly from your menubar (https://4gray.github.io/oversetter/)


zeke:
This looks cool. How does it work? Does it use Google or MS or something?

The description is great, but I think it could be a bit shorter and still clear. Maybe something like:

> Simple translation directly from your menubar.

or..

> Translate text directly from your menubar.

4gray:
Thank you! Yes, it uses Yandex Translate API. 
I have changed the description.

zeke:
👍

PR #222, opened 2017-07-11
Please consider adding to electron's showcase.

meekimobile:


zeke:
This looks good, but it seems you accidentally overwrote this project's README. Please undo that change and I'll ship it.

meekimobile:
I've reverted the change. Thanks.

PR #225, opened 2017-07-11
Add categories to apps

jkleinsc:
This PR adds categories to apps so that we can eventually show apps by category.

The following was also done:
* Apps that didn't have keywords should now have at least one keyword.
* Apps that were using the "Electron" keyword have had that keyword removed as it is unnecessary.


zeke:
Nice!

Can you add a test to human-data.js that ensures every app has a category, and that the category is one that we've defined/allowed?

We should probably also document the categories in https://github.com/electron/electron-apps/blob/master/contributing.md#adding-your-app

jkleinsc:
@zeke I added the tests, documentation and I also updated the wizard to ask for a category.

zeke:
Nice.

I created a followup issue at https://github.com/electron/electron.atom.io/issues/759

PR #230, opened 2017-07-20
add github-browser

tborychowski:


zeke:
Our legal team has advised us to disallow apps that are using the names of other companies or icons that we find too similar to the logos of other companies without verifying their permission to do so. We would still be happy to feature your app. If you are an author of this app and would like to see it on the site, please change the icon and update this PR to use the new icon.

Apologies for the inconvenience, and thanks for your understanding.

cc @lee-dohm

lee-dohm:
For future reference, with regard to GitHub's logos see the [GitHub logo guidelines](https://github.com/logos). As for the name of the app, it shouldn't start with the word "GitHub". In general, you may refer to GitHub in a relational phrase to say that the project is "compatible with", "on", or "for" GitHub, or something along those lines.

tborychowski:
understood, thanks!

zeke:
@tborychowski please do submit another PR if you get around to making those changes.

PR #231, opened 2017-07-20
Add Appium to apps

dpgraham:
(cc @jlipps)


apps/appium/appium.yml
```diff
@@ -0,0 +1,8 @@
+name: Appium
```
jlipps: I'd call this "Appium Desktop" instead of just "Appium"

zeke:
Thanks! I would suggest adding some keywords like hybrid, automation, webdriver, etc.

dpgraham:
Just added some more keywords. Thanks.

zeke:
👍

PR #241, opened 2017-07-31
Add Xuanxuan IM app.

catouse:


zeke:
Thanks, @catouse 

By default, your app is assumed to be designed for English speakers. If your app supports a different language (or multiple languages), please add a locales property that lists all locales supported.

See https://github.com/electron/electron-apps/blob/master/contributing.md#locales for more info.

Also `Communications` is not a valid category. See the list at https://github.com/electron/electron-apps/blob/master/contributing.md#adding-your-app-by-hand-

PR #242, opened 2017-07-31
Add Pepefe app

junwatu:


zeke:
Thanks, @junwatu. Can you add some keywords like `books`, `education`, `learning`, etc?

zeke:
Also please use screenshots that emphasize the app itself, not the entire desktop.

PR #243, opened 2017-07-31
Add devRantron to apps

RekkyRek:
devRantron: An unofficial open source client for devRant.io
Custom
Profile
Notif
View


zeke:
Thanks, @RekkyRek. `keywords` should actually be an array, not a string. I've added a test to catch that in the future: #244 

Would you mind updating your YML file? I would suggest adding more keywords while you're at it, to make it easier for folks to find your app when searching/filtering on the website.

PR #245, opened 2017-08-03
Add Ripplectron Application

devjin0617:
it's Ripple(Blockchain coin) Wallet Desktop client for vue-electron

start
wallet


zeke:
Thanks @devjin0617 

`utility` is not a valid category. See the list at https://github.com/electron/electron-apps/blob/master/contributing.md#adding-your-app-by-hand-

Also wondering if there's a way to improve the description so it doesn't need that parenthetical. Maybe something like "Desktop wallet for Ripple blockchain coins"... ?

devjin0617:
Thanks! I Updated description with category

zeke:
💰

PR #250, opened 2017-08-05
Add Tunlookup App

tun-application:


zeke:
I would suggest changing the description to something like "Translate selected text from your menu bar", or "Translate selected text with a keystroke", or similar depending on how your app works.

If the app interface is not in English, please also add a `locales` array. See https://github.com/electron/electron-apps/blob/master/contributing.md#locales

tun-application:
Thank for your suggestion, I has changed the description. 
And about language my app, only support english, so `locales ` is unnecessary. pls approve my app.
Thank a lot!

apps/tunlookup/tunlookup.yml
```diff
@@ -0,0 +1,10 @@
+name: Tunlookup
+description: 'Translate selected text with a keystroke'
+website: 'https://tun-application.github.io/'
+keywords:
+    - Tunlookup
+    - Translate selected text with a keystroke
```
zeke: These first two keywords can be removed, as they are the same as the name and description of the app.

zeke:
👍

PR #256, opened 2017-08-08
Add  Intu-Mind into electron/apps

IntuMind:
Intu-Mind is an idea organizer, is like a galaxy of ideas. Simple and easy interface for the user. Uses materialize.css and Electron js.
https://intumind.github.io/


zeke:
screen shot 2017-08-09 at 9 05 38 am

please use a lowercase slug and make sure your yaml file and icon name match the directory name.

PR #258, opened 2017-08-09
Add Source Me

theIYD:
Application Repository :- https://github.com/theIYD/source-me#readme


zeke:
Thanks. Can you add a more descriptive description? Maybe something like "Web development toolkit with fonts, icons, and more"

Keywords are recommended but if you don't want to add any, you can just omit the `keywords` property entirely.

theIYD:
Updated sir. :+1: :

zeke:
👌

PR #264, opened 2017-08-10
Add Storm, a file synchronization tool

payne8:
Overview of what Storm is:
http://www.stormfolder.com/

Sign up and or download for free here:
https://app.stormfolder.com

Storm will synchronize files stored in the Storm folder on your computer.

The synchronized folder is found here:
Mac: ~/Storm/
Windows: %Homefolder%\Storm\

Storm uses end-to-end encryption and perfect forward secrecy to transfer files. It also keeps a local copy of every version of a file in the case you need to restore from a previous version (previous versions are only stored when Storm downloads from other clients). Look for the previous versions at:
Mac: ~/Library/Application Support/Storm/versions/
Windows: %APPDATA%\Roaming\Storm\versions\

If you have questions please contact support@masonitestudios.com.

Also feel free to contact Mason Payne here:
GitHub: @payne8
Twitter: @emasonpayne

storm-login
storm-file-transfer


zeke:
Thanks @payne8. I would recommend using an icon with an opaque background. The entire image doesn't have to be opaque, but the itself icon should be.

You can also add these above screenshots to your metadata if you like. See details in https://github.com/electron/electron-apps/blob/master/contributing.md#adding-your-app-by-hand-

payne8:
@zeke for the icon do you mean adding a background color that makes the entire icon opaque rather than having the transparent portions between the separate colors? Let me know if the updated icon is more like what you meant.

Thanks for reviewing!

zeke:
👍

PR #267, opened 2017-08-11
Add Shopify to apps list

feens:


zeke:
Hi @feens. Does Shopify have a desktop app? I don't see a link to it on the homepage.

feens:
We do in the Mac App Store currently, Windows App Store is pending.  We haven't linked to it/them from our website yet.

zeke:
Got it. In that case can you link to the MAS website URL?

zeke:
(You can update the entry here when downloads are available for other platforms)

feens:
@zeke made that change

zeke:
👍

PR #271, opened 2017-08-15
Add TIM app

tborychowski:
Following up from the previous PR: https://github.com/electron/electron-apps/pull/230
I renamed the app and changed the icon :-)
I hope all is fine now!

@zeke @lee-dohm


zeke:
Looks good. Thanks for following up!

PR #280, opened 2017-08-24
add app "weweChat"

trazyn:


zeke:
Thanks! It looks like this app's UI is not in English. Can you add supported locale(s) to the metadata?

See https://github.com/electron/electron-apps/blob/master/contributing.md#locales

PR #288, opened 2017-09-07
adminScheduler

danieltoorani:
Scheduling for client/admin relationships


zeke:
Thanks for the PR!

It's not necessary to mention Electron in the description. I would recommend something like "Scheduling for client/admin relationships"

danieltoorani:
So Ive updated the description is there anything else I need to do?

zeke:
👍

danieltoorani:
Thank you!

PR #291, opened 2017-09-07
Twitch Desktop App

Acenth:
The Curse app became the Twitch Desktop a while back. Wanted to update the entry with latest information.


zeke:
Thanks!

I think the old description is better as it's concise and more clear to anyone unfamiliar with Twitch.

Acenth:
Changed!

zeke:
Nice! Thanks.

PR #293, opened 2017-09-08
add gHentai Downloader

dvgamer:
**gHentai Downloader**
manga queue downloader from e-hentai.org

image


zeke:
Thanks for your submission, but material of this nature is not appropriate for the electron apps collection.

PR #300, opened 2017-09-14
Rewrite local image urls in readmes

zeke:
This moves a little more work out of the Electron website code and into the build process for this module.


zeke:
PS this was built on top of the (unmerged) #299 branch. :)

PR #308, opened 2017-09-28
Add GitHoard app

jojobyte:
GitHoard makes it easy to quickly clone Git repositories.
https://github.com/jojobyte/githoard


zeke:
Thanks!

PR #311, opened 2017-10-03
Added NATTT app

TimoKorinth:
NATTT is Not Another Time Tracking Tool.
No timers you have to start and stop. No complex task creation dialogs. Just say what you are doing. It's that simple. Oh, and it's FREE.
[NATTT website](http://nattt.net/)
--
NATTT is an electron & cordova based application.


zeke:
> electron & cordova 

Interesting! Can you share more about how you're using the two in combination?

TimoKorinth:
Hey @zeke , sure. NATTT is basically an Angular application and I created it using the Angular CLI in the first place. I wanted it to be a web / Angular-first application and added integration with electron and cordova afterwards. The electron integration was easy, I just created a main.js file and added it to the package.json. The only thing I had to change was the default output folder (dist -> www).
The cordova integration was a bit more of a problem, but I managed it by creating a subfolder for the cordova stuff and created a symlink for the www folder. 
In order to get the default Angular routing to work, I also had to change the routing strategy to use the "old" hash strategy.
With this architecture I'm now able to develop my application like a normal Angular app and build & test it for electron & cordova with a simple build step.

zeke:
✨ Fantastic ✨ 

If you were to blog about this setup, I'm sure many people would be interested. We could also promote it from the Electron side as a way to build and package web applications for all platforms.

TimoKorinth:
Actually I already work on some blog posts around NATTT ;-)
I will let you know if the first (technical) one is ready and it would be awesome if you'll promote it. I already wrote a small one regarding my motivation to create this app: http://timokorinth.de/nattt-not-another-time-tracking-tool/.

PR #327, opened 2017-10-17
disable nylas

zeke:
Nylas is shutting down or something.

cc @bengotow


bengotow:
Yep - they've sunset Nylas Mail. I've added the Mailspring fork in #328!

PR #332, opened 2017-10-18
Fetch GitHub release data again

zeke:
This PR turns GitHub Releases and README fetching back on!

I updated it to use the [bottleneck](http://ghub.io/bottleneck) package, which gives us control over API request concurrency. It's currently set to `4` concurrent requests, and all 200ish releases and readmes download in under 20 seconds on my machine.

For apps that have a GitHub repo:

- `latestRelease` is the pure object we got from GitHub. No tampering. (cc @MarshallOfSound whose Fedora assets were getting stripped out)
- `readmeOriginal` is what we got from GitHub
- `readmeCleaned` has relative image URLs made absolute
- `latestReleaseFetchedAt` and `readmeFetchedAt` timestamps ensure we don't fetch too often: data is fetched at most once every 4 hours per app

cc @jkleinsc

### DIFF
--- a/lib/apps-with-github-repos.js
+++ b/lib/apps-with-github-repos.js
@@ -1,0 +1,11 @@
+const apps = require('./raw-app-list')()
+const parseGitUrl = require('github-url-to-object')
+
+module.exports = apps
+  .filter(app => {
+    // inherit repository from website if possible
+    if (!app.repository && parseGitUrl(app.website)) app.repository = app.website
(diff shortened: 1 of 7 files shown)


zeke:
Note that this is technically a breaking change as the app object structure has changed a bit, however I think it's safe to say that the new website is likely the only consumer of this part of the data. That said, I'm comfortable letting the Heroku bot ship this as a minor release.

PR #334, opened 2017-10-19
Accessible Color Warnings

zeke:
The [pick-a-good-color](https://github.com/zeke/pick-a-good-color) module will always pick an accessible color, but sometimes the color just doesn't look quite right.

To work around this, #325 enabled app developers to choose their own colors. This PR is a followup that ensures sure the human-provided colors are in fact accessible according to the [WCAG guidelines](https://www.w3.org/TR/WCAG/#visual-audio-contrast)..

When a color's contrast ratio is not high enough, this kind of error is produced:

```
AssertionError: google-play-music-desktop-player: contrast ratio too low for goodColorOnWhite. Try: #DE2811
```

cc @MarshallOfSound. I took the liberty of adding custom colors for GPMDP as a test case, based on the icon and the website.

cc @codebytere

### DIFF
--- a/apps/google-play-music-desktop-player/google-play-music-desktop-player.yml
+++ b/apps/google-play-music-desktop-player/google-play-music-desktop-player.yml
@@ -7,6 +7,8 @@ keywords:
     - GPMDP
 category: Music
 license: MIT
+goodColorOnWhite: '#DE2811'
+goodColorOnBlack: '#FFCA16'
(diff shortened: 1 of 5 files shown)


zeke:
This also updates the CONTRIBUTING doc to explain the contrast requirements and link to a site that makes it easy to pick accessible colors.

Also broke the doc up a bit with more headings so we can more easily deep-link to sections in the doc.

PR #345, opened 2017-10-22
Remove Front-End Box because it doesn't exist

noamtamim:
The app URL -- http://frontendbox.io/ -- is now just an ad dump.


zeke:
Thanks!

PR #361, opened 2017-11-11
add luna

rvpanoz:


zeke:
I would recommend adding a background to the "icon" portion of your icon image. That way it will display properly regardless of the background color it's displayed on.

rvpanoz:
Hey, thanks for your suggestion. I will fix it soon.

zeke:
nice!

rvpanoz:
@zeke thanks :)

PR #378, opened 2017-11-19
Added Synap (with screenshots as per suggestion)

jpgupta:


zeke:
Looks great! Just so you know you can always modify an existing pull request. No need to open aother one. See https://guides.github.com/introduction/flow/

PR #384, opened 2017-11-22
fix README.md MIT Link

kstar0714:

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -81,4 +81,4 @@ The bot can be extended to collect other useful data, especially if the app has 
 
 ## License
 
-[MIT](htps://github.com/electron/electron/blob/master/LICENSE)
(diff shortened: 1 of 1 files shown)


zeke:
I wonder if this repo should share a license with electron/electron, or have its own... 🤔 

Paging @lee-dohm.

lee-dohm:
@zeke Since this repo contains scripts and such too, you could probably go either way because you'll probably be using MIT here as well. Generally though since the initial copyright years are going to be different, I would lean toward having a separate `LICENSE` here.

zeke:
> I would lean toward having a separate LICENSE here.

I agree. @kstar0714 would be willing to add that?

kstar0714:
I`ll try.

zeke:
@kstar0714 I'm gonna close this but feel free to respond here if you want to pick this effort up again. Thanks.

PR #394, opened 2017-11-27
Added WWII Stats Viewer

ImNotLiam:
Added WWII Stats Viewer


Toinane:
Thank's for your change 😄 
For the name of your app, instead name: wwii-stats-viewer
write name: 'WWII Stats Viewer' isn't greatest?

PR #396, opened 2017-11-27
add Thomas app

andrepolischuk:
http://github.com/andrepolischuk/thomas

screenshot


andrepolischuk:
Oh, sorry. I thought that "at least 256" means icon can be large than 256 px.

Toinane:
Oh! Yeah you are right, I made a misunderstanding.. I should go to sleep ahah! Sorry :/

PR #400, opened 2017-11-29
add chronos-timetracker app

VladimirPal:


zeke:
CI is failing because the app icon a bit too big:

> expected 1067 to be below 1025

VladimirPal:
@zeke Thank you.

PR #419, opened 2017-12-10
Add WorkPuls to apps list

ipetrovic11:
We are SaaS startup called WorkPuls(https://www.workpuls.com).
We have desktop app which is one important part of our service which is tracking applications and websites usage and on which tasks or projects team is working.

alt text

You can download MacOS "binary" for testing purpose here:
http://cdn.workpuls.com/tmp/proof.zip
If you plan to run it you need to run it via CMD with at least one argument, best option is to use this argument "ac67a5c2-426a-4667-9b95-3818fa304c56"


ipetrovic11:
I was reading some previous pull request and as I understand these screenshots are required also as proof that it is electron app:

PR #435, opened 2017-12-18
I just added my app to electron-apps

MEGAMINDMK:


MEGAMINDMK:
please let me know if my app gets listed on electron apps
thnx
regards

PR #441, opened 2017-12-22
Pull request to add Mercury app to Electron's apps pool

Delni:
Mercury is a budget planner for personnal use


Delni:
The url is updated. I hope this one will work !

PR #451, opened 2017-12-31
Resolve relative links

MarshallOfSound:
Do what we do for images but for links as well 👍


zeke:
Fixes https://github.com/electron/electron-apps/issues/453

PR #481, opened 2018-01-26
Added Vocare HelpDesk

hammy2899:
Added Vocare HelpDesk the list of apps


hammy2899:
Here is the `package.json` showing it includes electron

image

and here is the install directory

image

---

As for the permission I currently work as a developer for Vocare and have had the approval to add the app to the list.

PR #489, opened 2018-02-03
fix broken link

alexander-shvets:
I've got ERR_CONNECTION_REFUSED on https://fastlaneapp.co/ but http worked fine.


alexander-shvets:
Hm...  http://fastlaneapp.co/ doesn't work too :(
It's Domain Suspension page.

PR #497, opened 2018-02-08
PhreshPlayer app added.

Phreshhh:


zeke:
Your app is live on the site now. Sorry for the delay! https://electronjs.org/apps

PR #498, opened 2018-02-09
Added Dotgrid

neauoire:
https://producthunt.com/dotgrid

preview


neauoire:
Also added Left.

https://www.producthunt.com/posts/left

preview

vanessayuenn:
Ooops, accidentally merged in all the commits from master when I was trying to rename the files 😓. But the tests are actually [expecting everything to be in lowercase with dash separator](https://github.com/electron/electron-apps/blob/master/contributing.md#adding-your-app-by-hand-). We should investigate why the error messages from CI were so unhelpful, though; it should have shown up like this:
```
> mocha --reporter min test/human-data.js && standard --fix

  10060 passing (974ms)
  1 failing

  1) human-submitted app data Left is in a directory whose name is lowercase with dashes as a delimiter:

      AssertionError: expected 'left' to equal 'Left'
      + expected - actual

      -left
      +Left
```

neauoire:
Mhm, it's strange, when I change the case of the yml file, it's not detected as a file change.

vanessayuenn:
@neauoire are you on a case insensitive system like mac os? If so, you'll need to use `git mv` to make git detect the change in casing.

neauoire:
I am! Okay, thanks :) I will give that a try.

vanessayuenn:
_bump_ :) @neauoire

neauoire:
AaaaaAahh @vanessayuenn!  I totally forgot it was still open.
Okay so I do I resolve this? I'm kind of a GitHub new and I'm seeing all these flags here. 
What should I do :')

neauoire:
Thank you so much 👍 
Let me know if you have ideas for Left and I will be happy to implement them!

PR #499, opened 2018-02-09
Changed category

nblackburn:
Audio notes would be better suited in the music category


ckerr:
Thanks for this patch!

@nblackburn, I think it's reasonable in either category, so I'm going to leave it in the one chosen by the program authors and close this PR.

If you are on the AudioNodes team and still want this to be changed, please @ me in a followup comment and I'll merge the PR.

nblackburn:
@ckerr Fair enough 👍

PR #512, opened 2018-02-17
Add Image Shrinker

stefansl:
App: Image Shrinker

Description: 
Minify your images and graphics with just one drop. Autorenamed in the same place where it comes from. Immediately

Website: https://image-shrinker.com

Screenshot:
Screenshot Imageshrinker


zeke:
Nice work! Looks like a useful app.

PR #517, opened 2018-02-21
LaTeX and WYSIWYG editor written with electronjs

fetacore:
It would be an honor for me if you included my app in your list. It is my first project on electronjs and Github and I already use it myself for my academic writing and I believe that others will benefit from this friendly approach towards writing LaTeX documents. The simple part is fully functional and will see many additional features in the future. Automatic updates work on both provided binaries. Thank you!
[Here](https://github.com/fetacore/Infinitex) is the URL of the repo.
A screen recording gif of the app in action:
alt start screenshot


zeke:
Thanks!

PR #524, opened 2018-02-25
Re-enable Zeplin

merih:
Re-enable Zeplin per https://github.com/electron/electron-apps/issues/272#issuecomment-368147294


zeke:
👍

PR #536, opened 2018-03-01
Delete tweetman.yml

stephanvierkant:
Website is offline.


vanessayuenn:
@stephanvierkant we actually just "disable" apps instead of removing them entirely from the registry. Could you please modify this PR according to [the contributing guide](https://github.com/electron/electron-apps/blob/master/contributing.md#removing-or-disabling-apps)?

PR #548, opened 2018-03-06
Add Keyfant

HalilBu:
URL: http://keyfant.de

linux1
macos1


HalilBu:
Built with Electron evidence:
evidence_electron

PR #549, opened 2018-03-06
Add Buckets budgeting app

iffy:
Buckets lets you do beautiful, private budgeting.

- https://www.budgetwithbuckets.com/

Here's a few screenshots:

screen shot 2018-03-06 at 12 51 28 pm

screen shot 2018-03-06 at 12 52 23 pm

And proof that it's an Electron app (since it's closed-source):

screen shot 2018-03-06 at 12 55 02 pm


iffy:
Oh, is the screenshot at the end not enough?  Let's see...  is this enough?

screen shot 2018-03-07 at 8 59 41 am

PR #550, opened 2018-03-06
Add FreeMAN

matthew-matvei:
[FreeMAN](https://github.com/matthew-matvei/freeman) is a minimalistic, keyboard-focused file manager for power users with an integrated terminal for those hard-to-reach actions.


matthew-matvei:
My pleasure

PR #562, opened 2018-03-12
Add Container PS

Toinane:
- Add a little app 'Container PS' that list your Docker containers and allow you to manage them.


Toinane:
Oh, how you can see the image is corrupt?
But no problem, I'll make a new variant icon :)

Toinane:
Here we go :) It's updated

PR #566, opened 2018-03-16
add paymo time tracker

alexdrimbe:


alexdrimbe:
Here is the evidence that this is built with Electron:

screen shot 2018-03-16 at 3 16 12 pm

Thank you!

PR #573, opened 2018-03-21
add crypto-bot

Voyager-One:


apps/crypto-bot/crypto-bot.yml
```diff
@@ -0,0 +1,12 @@
+name: Crypto-Bot
+description: 'app for encryption and decryption'
```
zeke: No need to put "app" in the description. Also is it for encrypting files? Messages across a network?

Voyager-One: no network request. It works locally

zeke: Got it. Please update the description so this is clear to everyone.

Voyager-One:
####it's like this : )
image
image

PR #591, opened 2018-03-30
bump minimum icon size to 256px

zeke:
The documented minimum icon size was 256, but it's actually 128.

Unfortunately there are 67 apps with icons below 256, so we can't easily change that threshold:

https://gist.github.com/zeke/d69001334e91a7122a8d2ca20ab42efa

Tho we could require _newly added apps_ to meet the 256 requirement, and make exceptions for older apps. 🤔 

@vanessayuenn


ckerr:
+1  on keeping the documented level as-is at 256, and requiring it for new apps

zeke:
Okay revamped this PR. Docs remain the same. New minimum is 256px, but 67 existing apps with smaller icons are grandfathered.

test/human-data.js
```diff
@@ -141,11 +142,11 @@ describe('human-submitted app data', () => {
...
         })
 
-        it('is at least 128px x 128px', function () {
+        it('is at least 256px x 256px (or 128px x 128px for grandfathered apps)', function () {
           if (!fs.existsSync(iconPath)) return this.skip()
-
           const dimensions = imageSize(iconPath)
-          expect(dimensions.width).to.be.above(127)
+          const min = grandfatheredSlugs.includes(slug) ? 128 : 256
+          expect(dimensions.width).to.be.above(min - 1)
```
vanessayuenn: I think `expect(dimensions.width).to.be.at.least(min)` will be easier for humans to understand.

zeke: ooh didn't know about `at.least` 👍

zeke:
Build was failing because Node 4 doesn't have `Array.includes`

vanessayuenn:
@zeke any particular reason we're running builds in multiple node version? Node 4.x is also slated for end-of-life at the end of this month.

zeke:
We used to support Node 4 everywhere because GitHub's CI was stuck on that version. I'll open a new PR to remove it from the Travis config.

PR #616, opened 2018-04-18
Added wallpaper viewer

ImNotLiam:
# Wallpaper Viewer
image viewer that allows you download wallpapers and set them as your desktop background.


zeke:
Thanks!

PR #626, opened 2018-04-27
add an app e-tools

Suremotoo:
Add an app,e-tools.It's a tool that doesn't need networking.


Suremotoo:
Add an app,e-tools.A web application that integrates a variety of tools and does not require networking，for platforms of OS X、Windows and Linux.

zeke:
How about something like "A toolbox for developers. Color picker, code formatter, and more."

Can you also add a [`locales`](https://github.com/electron/apps/blob/master/contributing.md#locales) property? 👍

PR #633, opened 2018-05-10
Add Cleaver

ashokgelal:


ashokgelal:
Here is a screenshot that shows app.asar and electron.asar from Cleaver. I hope this is enough.

PR #644, opened 2018-05-14
Add "MMR Calculator" to app list

AustinLeath:
https://github.com/AustinLeath/mmrcalculator
https://www.r6rc.com/mmrcalculator

I believe that The Rainbow Six: Siege | MMR Calculator should be added to the app list because of the amount of time that I have dedicated to this project.

The Rainbow Six: Siege | MMR Calculator uses a users current MMR and ELO to display how many matches the user has to win or lose to reach their goal.

The Rainbow Six: Siege | MMR Calculator is in no way shape or form affiliated with Ubisoft and its partners. Any references to "Rainbow Six: Siege" and its ranking system logos and images are registered trademarks of Ubisoft.


AustinLeath:
Thank you for considering my submission, let me know what I need to improve if my submission does not pass

PR #656, opened 2018-05-22
Add Winds 2.0

nparsons08:
Add https://getstream.io/winds

https://i.imgur.com/n2eKwPf.gif


zeke:
This looks great! But your YAML file needs to have the same name as your app, i.e. `winds.yml`

Also you should consider adding some keywords so your apps shows up for more searches:

```yml
keywords:
  - feed
  - rss
  - podcast
  - audio
  - syndication
```

nparsons08:
@zeke All set! Thanks for the recommendation!

zeke:
Thanks!

PR #657, opened 2018-05-22
Update Diffuse icon & info

icidasset:
Hey folks,

Would it be possible to update the icon for my app?
Thanks!

https://diffuse.sh/


zeke:
Looks good, though the `diffuse-icon-256.png` file is not necessary and will be ignore. Can you remove it?

icidasset:
Thanks Zeke, I removed the file.

PR #668, opened 2018-06-06
Adding Sample Video in the description of hyper as the previous GIF was a 404 content.

Pravandan:
The previous link was not working so therefore I took the sample video from hyper.is and inserted in the README


amorist:
#669

PR #674, opened 2018-06-11
Add Voice Notifies to app list

usergit:
here is a link to my app http://voicenotifies.com I used electron 1.7.8. To verify it was built using electron, you can go to the about section and see the electron build number. No company logo or trademark used. It also has been more than a month since I created the application

Thanks for the addition


apps/voice-notifies/voice-notifies.yml
```diff
@@ -0,0 +1,15 @@
+name: 'Voice Notifies'
+description: 'Google Voice compatible desktop app'
```
MarshallOfSound: This should be the way you had it one commit ago. 

I.e. `XXXX compatible with Google Voice`

usergit:
I prefer the other name but "Desktop app compatible with Google Voice" will do

usergit:
I've made the change as requested, hopefully this will resolve the issue

PR #687, opened 2018-06-19
trying to add RawFileCleaner to the Electron Apps

ChristianStrassmayr:
RawFileCleaner is a tool to manage photo archives.

It's pretty simple. Most new cameras create a compressed and a raw file when you take a photo, but when you delete compressed photos you don't like, the raw files which belong to the deleted compressed file are not deleted automatically. RawFileCleaner deletes the raw files which belong to the compressed files you already deleted.


codebytere:
We haven't received a follow-up for the changes requested above so i'm going to go ahead and close this. You're more than welcome to open another one with the correct information 😄

PR #688, opened 2018-06-21
Updated octopi-init.yml

OutsourcedGuru:
Unfortunately, the yaml file is a snapshot in time and doesn't reflect the latest/greatest download link(s). It was necessary to create a separate `electron_README.md` as an endpoint.

### DIFF
--- a/apps/octopi-init/octopi-init.yml
+++ b/apps/octopi-init/octopi-init.yml
@@ -2,7 +2,7 @@ name: octopi-init
 description: 'A Node.js & Electron application for easily configuring your freshly-imaged OctoPi microSD adapter'
 website: 'https://github.com/OutsourcedGuru/octopi-init'
 category: Utilities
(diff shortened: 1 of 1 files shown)


OutsourcedGuru:
@ckerr Unfortunately, this still doesn't produce the expected result. I'm hoping that the octopi.yml version displays the `electron-README.md` file on the repository instead of the default `README.md`. In this way, it will be a stub which includes a link to take you to the repository where you may find the (maintained) linked buttons to download.

What's required for this to happen?

OutsourcedGuru:
Are you sure?  I'm seeing this on your site:

screen shot 2018-06-21 at 12 26 52 pm

...which appears to include my attempt to make this work (note the `electron-README.md` referenced which is new).

And it's still got the download buttons:

screen shot 2018-06-21 at 12 28 02 pm

I should shut up and trust you, right?

PR #689, opened 2018-06-21
Add Sharp Tune

MD-AZMAL:
Add my music player to apps


apps/sharp-tune/sharp-tune.yml
```diff
@@ -0,0 +1,11 @@
+name: Sharp-Tune
+description: 'Lightweight Cross-platform music player build upon the node using the electron framework.'
```
zeke: No need to mention node and electron in the description. How about just `Lightweight cross-platform music player`?

MD-AZMAL: sounds great!

MD-AZMAL: Please check i have made the necessary changes .......... thanks for the advice

MD-AZMAL: @zeke i have updated the screenshot for the app as well and added a few more keywords have a look!

zeke:
Thanks!

Issue #602, opened 2018-04-06
»Thunder« app outdated?


jancborchardt:
The app [Thunder](https://electronjs.org/apps/thunder) seems to be outdated as it’s unreachable and I couldn’t find further info anywhere.

According to whois.net, the domain trollswebapp.com expired 2 days ago. I couldn’t find the icon or description anywhere else either.

cc @davidsolorzano20 who added it with https://github.com/electron/apps/pull/214

AustinLeath:
must definitely be deprecated.

jancborchardt:
Pull request to disable it at https://github.com/electron/apps/pull/694

PR #694, opened 2018-06-23
Disable Thunder app, ref #602

jancborchardt:
As mentioned in https://github.com/electron/apps/issues/602:

> The app [Thunder](https://electronjs.org/apps/thunder) seems to be outdated as it’s unreachable and I couldn’t find further info anywhere.
> 
> According to whois.net, the domain trollswebapp.com expired 2 days ago. I couldn’t find the icon or description anywhere else either.
> 
> cc @davidsolorzano20 who added it with https://github.com/electron/apps/pull/214

### DIFF
--- a/apps/thunder/thunder.yml
+++ b/apps/thunder/thunder.yml
@@ -6,3 +6,4 @@ keywords:
 locales:
     - es-NI
 category: 'Social Networking'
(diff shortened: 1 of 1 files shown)


zeke:
Thanks, @jancborchardt

PR #706, opened 2018-07-09
Add 'kalk' and 'clippo'

monsterkodi:
Clipboard Tool
https://github.com/monsterkodi/clippo

clippo
clippo

A(nother :) Simple Calculator
https://github.com/monsterkodi/kalk

kalk


apps/clippo/clippo.yml
```diff
@@ -0,0 +1,18 @@
+name: kalk
```
zeke: This should be `Clippo`, right?

monsterkodi: Hi zeke, 

not really, I like it lower case, but if necessary I can live with the upper case as well.

monsterkodi: Oops. sorry. Misunderstood your question. Yes, in the clippo.yaml it should be clippo instead of kalk.
I will fix that.

monsterkodi: Ok, should be fixed now. 
Thanks for the hint.

apps/clippo/clippo.yml
```diff
@@ -0,0 +1,18 @@
...
+screenshots:
+  -
+    imageUrl: 'https://raw.githubusercontent.com/monsterkodi/clippo/master/img/readme.png'
+    caption: 'windows'
+    imageLink: 'https://github.com/monsterkodi/clippo'
+  -
+    imageUrl: 'https://raw.githubusercontent.com/monsterkodi/clippo/master/img/shot.png'
+    caption: 'mac'
+    imageLink: 'https://github.com/monsterkodi/clippo'
+
```
codebytere: remove this newline

PR #716, opened 2018-07-16
Update one-left.yml

drscaon:


drscaon:
How long does it take to update the download version which is shown at https://electronjs.org/apps/one-left ? Is it automatic? What is the frequency of updates? I've got a disparity between the github's releases. Thank you for the attention.

PR #725, opened 2018-07-20
A simple application launcher

monsterkodi:
https://github.com/monsterkodi/kappo

screenshot

screenshot


monsterkodi:
thanks for including it!

PR #747, opened 2018-08-09
Added sFTP Client

dannypbroadbent:
**sFTP Client**
_Transfer files with FTP, FTPIS, FTPES and SFTP (Including SSH Terminal)_

Built in AngularJS and available to macOS, Windows and Linux.

**Download**
https://www.sftpclient.io/download

**Screenshot**
Light / Dark Theme


dannypbroadbent:
Hello @ckerr,

I've amended the link to the direct download page.

We've also been known as sFTP Client for the past 4 years when we launched in the Google Chrome Webstore, and recently we used NW.js, however we prefer the Electron framework.

Our app provides SFTP protocol services, hence `sftp` being within the keywords as you know its a way for people to find our application.

We're also a fully registered business under sFTP Client Ltd.

In relation to point 2, if you download the mac / linux version and view the contents of the app file and you can then certainly see we use Electron.

Best Regards
Danny

dannypbroadbent:
screen shot 2018-08-09 at 22 33 09

ckerr:
@dannypbroadbent, thanks for the proof-of-electron and for the project info. I was thinking more of the sftp app itself, but it's a fair point that, since the protocol itself is also named sftp...

Last step, do you have a larger icon that can be used to make CI happy? :smile:

dannypbroadbent:
Sorry was following Icons of others, i.e. Slack, we have a 1024 x 1024

dannypbroadbent:
Updated with a 512 otherwise 1024 is going to a larger file size, you also probably don't have a need for one of that size anyway....

PR #766, opened 2018-08-15
feat: Add School Timetable app

KDani-99:
https://github.com/KDani-99/timetable

It is a simple and very useful timetable application that allowes people to add/edit/remove lessons/activities. This would help many students/people to prepare for the next day(s).


KDani-99:
Why did I do wrong?

KDani-99:
Oh I think we found out what we did wrong 🙃

KDani-99:
Is it ok now?

PR #773, opened 2018-08-18
refactor: recompute colors only when needed

ckerr:
Only generate complimentary colors for an icon file when that file changes.

 - [x] Generating these colors are expensive, so keep revHashes of the icon files and check that when updating to see if the previously-stored value can be reused.
 - [x] Refactor the complimentary color generator in such a way that the endpoints can be better controlled for testing
 - [x] Add unit tests for the color generator

Nota bene: I'm using this PR as a testing ground for a lot of JS features / node modules I haven't played with before. Corrections on incorrect / non-idiomatic / double-plus-ungood code is welcomed :+1:


test/colors-spec.js
```diff
@@ -0,0 +1,138 @@
...
+
+const Colors = require('../lib/colors.js')
+
+describe('colors', function () {
+  let testDir
+  const slugsAndIconPaths = []
+
+  before(async function () {
+    console.info = () => {}
+    console.error = () => {}
```
zeke: Is the idea here that logging should be visible when the build is run but not the tests? I get that, but I think `console.error` output could be helpful for diagnosis when this test fails.

test/colors-spec.js
```diff
@@ -0,0 +1,138 @@
...
+    const unchangedEntry = entries[1]
+    changedEntry.iconPath = unchangedEntry.iconPath
+    const newColors = await Colors.getColors(entries, oldColors, testDir)
+
+    // the revHash on the unchanged entry should be unchanged
+    expect(newColors)
+      .property(unchangedEntry.slug).to.deep.contain(oldColors[unchangedEntry.slug])
+
+    // the revHash on the changed entry should be different
+    expect(newColors)
```
zeke: I don't think the indentation is necessary for assertion chaining; left justified would probably read better.

lib/colors.js
```diff
@@ -0,0 +1,95 @@
...
+const colorConvert = require('color-convert')
+const getImageColors = require('get-image-colors')
+const mime = require('mime-types')
+const path = require('path')
+const pickAGoodColor = require('pick-a-good-color')
+const revHash = require('rev-hash')
+const stringify = require('json-stable-stringify')
+
+/**
+ * Generates complimentary colors for an image.
```
zeke: `compl(i|e)mentary` is not really the right term here; the color list generated by `get-image-colors` is a set of the "dominant" or "quantized" colors in an image.

test/colors-spec.js
MarshallOfSound: Also this means that any legit console.errors will be sent to the void

ckerr: :+1:

zeke:
Looks like the build is now failing in a few places:


```
Error processing /tmp/colors-specufxfHL/white.png { Error: EACCES: permission denied, open '/tmp/colors-specufxfHL/white.png'
```

```
Error processing /tmp/colors-specufxfHL/hello.png Error: Invalid file signature
```

```
standard: Use JavaScript Standard Style (https://standardjs.com)
  /home/travis/build/electron/apps/test/colors-spec.js:9:7: 'sinonChai' is assigned a value but never used.
  /home/travis/build/electron/apps/test/colors-spec.js:104:5: Expected an assignment or function call and instead saw an expression.
  /home/travis/build/electron/apps/test/colors-spec.js:123:5: Expected an assignment or function call and instead saw an expression.
```

ckerr:
@zeke sorry about the delay on this -- the issues you found last week are resolved now.

zeke:
🚀

PR #777, opened 2018-08-24
feat: Add Vega Clipboard

aleskinmk:
Most easy to use clipboard manager.
vega clipboard manager

This is our site http://thevega.pro/


aleskinmk:
Sorry,
Will it be enough?
proof 1

PR #778, opened 2018-08-26
feat: add Cycle app

joshhowen:


joshhowen:
@ckerr no problem at all - here is the screenshot I used when Cycle was last accepted into the list:

cycle

ckerr:
Ah, looks like this got superceded by 21af3e94ba8fe9689a66ceb327d8faa8187ddad7 and then never got closed. Closing now for that reason.

If this is in error, please do create a new PR so that it'll show up on the maintainers' radar. Thank you!

PR #791, opened 2018-09-05
chore: update username of Museeks maintainer

martpie:
I recently changed my username on Github, because GH is smart and forward URLs nicely, it's fine, but making the change just in case :)


martpie:
Fair point indeed, done.

Issue #804, opened 2018-09-12
tests sometime fail in `expected {} to have key 'white'`


ckerr:
[Example](https://github.com/electron/apps/pull/803):

```
> mocha --reporter min test/human-data.js test/colors-spec.js && standard --fix

calculating good colors for white
calculating good colors for black
calculating good colors for white
calculating good colors for black
calculating good colors for white
calculating good colors for black
calculating good colors for white

  12967 passing (4s)
  1 failing

  1) colors should create entries with the expected properties:

      AssertionError: expected {} to have key 'white'
      + expected - actual

      -[]
      +[
      +  "white"
      +]
      
      at Context.it (test/colors-spec.js:57:13)
      at <anonymous>
```

PR #817, opened 2018-09-20
chore: update alva icon and screenshots

LKuechler:
Hi there,
over time we changed a bit on our website and with that the links to the screenshots.
Because of that i took the time to update our description as well.

Thanks again for this overview :)


ckerr:
https://github.com/electron/apps/issues/804

ckerr:
This build failure is due to an error in our tests (tracked in #804) rather than an error in this PR.

PR #832, opened 2018-10-01
feat: add Skutti app

JohannBlake:
Submitting my app to appear in Electron's app store:

[https://electronjs.org/apps](https://electronjs.org/apps)

**Proof It's An Electron App**

1. Download the app from [https://www.skutti.com](https://www.skutti.com)
2. Install the app. Further steps here assume you're installing the Mac version. If installing a Windows version, the location of the app is different.
3. Under Applications, scroll to the app "skutti".
4. Right click the app and select "Show Package Contents"
5. Navigate to the Resources folder. The **electron.asar** file is present.
6. For further proof, use the npm module **asar** and run the following command on the app.asar file which is also located in the Resources folder: ```asar extract app.asar skutti-code```
7. Navigate into the extracted source code located in the skutti-code folder that got created.
8. Open up the package.json file and check the dependencies. They will contain the electron modules.


JohannBlake:
Why did you change the title to feature?? I'm reporting a bug.

ckerr:
> @ckerr Why did you change the title to feature?? I'm reporting a bug.

I'm not sure I understand your question. The PR needed a semantic commit message to satisfy the [semantic release bot](https://github.com/apps/semantic-pull-requests) happy. Since this is a new entry in the Apps list, `feature` seemed appropriate. Am I misunderstanding your question?

However I should've read the bots' output more carefully as there was a second error with the app icon. Do you have one that's at least 256 pixels x 256 pixels that could be added to the repository?

PR #835, opened 2018-10-02
fix: bump nan dependency

ckerr:
This was required for me to `npm install` successfully on v10.11.0.Otherwise I hit a "no member named 'ForceSet' in 'v8::Object'" [error](https://github.com/nodejs/nan/issues/766) in nan (a subdependency of sharp) which is fixed by bumping the sharp dependency.


ckerr:
Well I'm using Node 10 on my machine, but I don't know if it needs to be a requirement...

PR #837, opened 2018-10-02
More readable test warnings

ckerr:
Try to make test output more readable on failure.

 - try to make the icon-must-be-within-these-dimensions error message more readable
 - capture & test the `console.info()` messages from the complimentary color generation pass, since that fills up the console with noise that takes focus away from error messages

### DIFF
--- a/test/colors-spec.js
+++ b/test/colors-spec.js
@@ -15,7 +15,9 @@ const expect = chai.expect
 const Colors = require('../lib/colors.js')
 
 describe('colors', function () {
+  let infos
   let errors
+  let consoleInfo
   let consoleError
   let testDir
   const slugsAndIconPaths = []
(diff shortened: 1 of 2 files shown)


test/colors-spec.js
```diff
@@ -43,10 +45,15 @@ describe('colors', function () {
     errors = []
     consoleError = console.error
     console.error = (...args) => errors.push(args)
+
+    infos = []
+    consoleInfo = console.info
+    console.info = (...args) => infos.push(args)
```
MarshallOfSound: This may be overkill for this PR but there is a common module that people use to follow this kind of pattern called [`sinon`](https://npmjs.com/sinon)

```js
const sinon = require('sinon')

sinon.stub(console, 'info') // Stubs the `info` method of `console`
console.info('foo') // Does not log
expect(console.info.callCount).to.equal(1) // You can do assertions on calls
expect(console.info.firstCall.args).to.deep.equal(['foo'])

console.info.restore() // Restores original implementation
```

PR #852, opened 2018-10-12
feat: Add Qawl to app list

mr-islam:
# Qawl 

[**Homepage**](https://qawl.navedislam.com/) 

[**Repo**](https://github.com/mr-islam/qawl) 

This app would be really helpful because it would allow people to have good page-setting, quick navigation, dynamic links to references, dark theme and so on — a major upgrade from the PDF scans people use currently.

Screenshots:

main

dark theme


mr-islam:
Hi, some help with this build error would be really appreciated:

```
12736 passing (4s)

  2 failing


  1) colors should create entries with the expected properties:


      AssertionError: expected {} to have key 'white'

      + expected - actual


      -[]

      +[

      +  "white"

      +]

      

      at Context.it (test/colors-spec.js:64:13)

      at <anonymous>


  2) colors should add an entry when a new app is added:

     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

Is this an error with the colors in the .png file? The first build failed because it was too large, but the second build with a 512px icon failed too

mr-islam:
This PR is done from my side now!

PR #854, opened 2018-10-14
feat: new app clipboard-manager-electron

dotenorio:
See more in https://github.com/dotenorio/clipboard-manager-electron


dotenorio:
Thanks @btargac

PR #859, opened 2018-10-19
feat: add YTMDesktop App

adlerluiz:
https://ytmdesktop.app
main 1


ckerr:
Renaming the app from `YouTube-Music-Desktop-App` to `YouTube Music Desktop App` doesn't address the question of using another company's trademarked name. Many people reading `YouTube Music Desktop App` would think it was software made by YouTube.

What I had in mind was something more like `$Name Desktop Music compatible with YouTube` where $Name was something of your own creation.

adlerluiz:
Hi @ckerr I sent it so because I saw that there are some apps with this same type of name as [name: 'Google Play Music Desktop Player'], [name: Youtube-mp3], [name: 'YouTube To MP3']. Was there any change in the submission guidelines after they submitted?

ckerr:
Yes. For example, "Google Play Music Desktop Player" seems the most similar to your app's name in that it is not using any of the "for Google Play" or "compatible with Google Play" style naming. It predates the current guidelines. I would not have approved it with that name under the current guidelines and question whether it should be kept at all under that name.

adlerluiz:
I understood, thanks for the explanation and for the approval =)

PR #863, opened 2018-10-21
add DN Tool Container App

como65416:
## DN Tool Container

An application can install many convenient tools for platforms of OS X 、Windows and Linux.

It can get more and more convenience tool in the future and no need reinstall Application.

### Repo:

https://github.com/como65416/DnToolContainer

### Feature:

- one click install and uninstall new package
- package extendable


como65416:
....

ckerr:
>

PR #866, opened 2018-10-22
chore: dependency bumps to fix npm audit warnings

ckerr:
Changes:

1. Ran `npm audit fix` to pick up the automatically-fixable packages

2. Manually updated the `GitHub` dependency to `@octokit/rest`

3. Manually bumped the dependency for our `check-for-leaks`

4. Confirmed that `npm run build` and `npm test` both Work for Me


lib/github.js
```diff
@@ -1,6 +1,6 @@
 require('dotenv-safe').load()
 
-const Github = require('github')
+const Github = require('@octokit/rest')
```
MarshallOfSound: This is a big bump and probably isn't 100% API compatible,  No CI ran on this build so would want to at least run CI and check the main scripts work with this bump before merging

ckerr: The maintainers _claim_ that it's API compatible, for what that's worth: https://github.com/octokit/rest.js/releases/tag/v14.0.0

> To upgrade
>
> npm install @octokit/rest
> npm uninstall github
> replace require('github') with require('@octokit/rest')
> That’s it, there are no functional changes in the public APIs compared to github@13.

That said, it's not clear to me why no CI ran on this PR...?

MarshallOfSound: @ckerr Yeah, webhooks aren't firing.  They should turn them back on soon then a quick force push should deal with it 😀

ckerr: @MarshallOfSound passing now

MarshallOfSound: There's a single breaking change from 14 -> 15 but I don't think that it affects us here

> Before, methods like octokit.issues.deleteLabel() returned an empty string for {data}. They now correctly return undefined

PR #868, opened 2018-10-23
feat: Updated Unicode Plus description

tonton-pixel:
Updated description and keywords in unicode-plus.yml


ckerr:
@tonton-pixel please resolve the conflicts so that this PR can be merged

PR #936, opened 2019-01-12
fix: clipboard memo : update website and img url

fabrom:


ckerr:
@fabrom could you fix those two conflicts?

PR #942, opened 2019-01-18
test: change mocha reporter from spec to min

HashimotoYT:
Changed mocha reporter from spec to min for show only errors.


MarshallOfSound:
This was a delibrate change I made while cleaning up and fixing test flakes to better show tests as they run / hang.  It's not much extra output but greatly helps with debugging

HashimotoYT:
I add `test:debug` command for debugging tests with spec reporter.

PR #963, opened 2019-02-05
feat: add agantty app

FlorianB385:
Our Desktop client for Linux, macOS, and Windows made with Electron.

Bildschirmfoto 2019-03-12 um 12 48 45


FlorianB385:
@ckerr i've changed the description of the pull request. is this so ok?

PR #986, opened 2019-03-04
feat: Change Name Of sticky-notes to stikynotes

bekalshenoy:
Change Name Of sticky-notes to stikynotes also change its content


bekalshenoy:
How is this build is failing? What's the problem

PR #996, opened 2019-03-12
 feat: Add 'stacker' to app list

lopidio:
Adding [stacker](https://lopidio.github.io/stacker) to the app list

stacker-electron-app
ignored-tests


lopidio:
Thank you!

PR #1001, opened 2019-03-13
feat: add "Saadhn" app

arkokoley:
I have been working on creating a desktop version of the JioSaavn Music Web App which supports native media buttons like play and pause etc. 

Screenshot

I had previously submitted this project at #973. Resubmitting after changing app name.


arkokoley:
@ckerr when would this go live on the website?

PR #1003, opened 2019-03-14
rename "email-securely-app" to "ElectronMail"

vladimiry:
The project has been recently renamed https://github.com/vladimiry/ElectronMail/releases/tag/v3.0.0


vladimiry:
I was not aware of such guideline. So closing the PR.

vladimiry:
@ckerr btw with help of http://tmsearch.uspto.gov/bin/gate.exe?f=tess&state=4803:nqqio.1.1 I can easily locate dozens of apps listed on https://electronjs.org/apps directory names of which include registered trademarks. Does it mean in all such cases evidence of trademark possession / prior owners permission was provided before pull request merging happened? Or is the restriction applied only selectively for Electron trademark? It would be good to know the answer for the future if I decide publishing new apps to the directory.

ckerr:
From https://github.com/electron/apps/blob/master/contributing.md#company-logos-and-names:

> While some existing apps in the collection predate this rule and have been grandfathered in, we will not accept any apps that do not follow this rule going forward.

PR #1008, opened 2019-03-20
feat: yml website update

kimyearho:


ckerr:
Please ignore my previous comment. I didn't understand that `metube-electron` was already in the repo -- hence the conflicts, I guess -- predating the contributor guidelines rule changes.

apps/metube-electron/metube-electron.yml
```diff
@@ -0,0 +1,12 @@
+name: Metube-electron
```
ckerr: As per https://github.com/electron/apps/blob/master/contributing.md#company-logos-and-names --

If GitHub has given you permission to use the name 'Electron' in your appname, please provide that information in the PR so that the app's name can be used as-is. Otherwise,

>

HashimotoYT:
Closing due to inactivity. Please open new PR if to add your application.

PR #1016, opened 2019-03-27
feat: add 'code-sketch' app

keelii:
add code-sketch app


apps/code-sketch/code-sketch.yml
```diff
@@ -0,0 +1,28 @@
+name: 'Code Sketch'
+description: 'First place where the code was written...'
```
ckerr: The App gallery's pretty loose on what descriptions are allowed; but even so, this description doesn't explain what the app is at all. Could you make this more descriptive of what the app does / what it's used for?

keelii: OK，let me think about another desc 🤔

PR #1018, opened 2019-03-28
feat: Cass

aAmitSengar:
this is the init commit and useful lib for Cassandra developers


aAmitSengar:
yes i already applied with Apache 
not a problem i will wait.

Thanks for quick response :+1:

aAmitSengar:
@ckerr please let me know if anything is required and Cassandra people told 

 **ME:**  Can i use cassandra GUI name as title of project?
**They: **   You can use the XXX for Apache Cassandra form of naming is you wish.

**ME:**  Can i use cassandra logo image?
**They: **  You can't use the original logo. You may be able to use a logo derived
from the original. We (trademarks@apache.org) would need to see a
proposed logo to provide a definitive answer.



so i accepted all conditions Please check and let me know.

PR #1024, opened 2019-04-05
feat: Added biblio app

JoshuaManuel:
Just a simple app to make bibliography citations


JoshuaManuel:
Biblio dynamically creates the citations you need as you enter in more information about your source

biblio in action

PR #1032, opened 2019-04-13
feat: add Knowte to list of apps

digimezzo:
Hi,

This is my new app "Knowte" which makes notes taking, organizing and searching very easy. it runs on Linux and Windows. This is the repository: [https://github.com/digimezzo/knowte-electron](https://github.com/digimezzo/knowte-electron)

image

Thank you for considering this submission.

Regards,

Raphaël


apps/knowte/knowte.yml
```diff
@@ -0,0 +1,8 @@
+name: Knowte
+description: 'Cross platform note taking application'
+website: 'https://www.digimezzo.com'
+category: Productivity
+repository: 'https://github.com/digimezzo/knowte-electron'
+keywords:
+    - 'note notes text organize search quick'
```
ckerr: I think you mean these to be separate keywords?

```suggestion
    - note
    - notes
    - text
    - organize
    - search
    - quick
```

digimezzo: Yes indeed

digimezzo:
My bad! I reverted the changes to package-lock.josn. Hopefully that does the trick.

PR #1046, opened 2019-04-27
feat: Playbox - an audio playback application.

cjdenio:
Hello!

Playbox is an application for playing back audio, such as sound effects and music. What sets it apart from other music players is that it's built for professional productions, such as theatre. This is my first full Electron app, so it's far from completed. (Also, I'm only 15) However, it's at the point where it's quite usable by anyone, even beginners. I have binaries for Windows, MacOS, and Linux [here](https://github.com/cjdenio/playbox/releases).

Thank you for your time, and have a great day!

Playbox screenshot

I'm really sorry if I messed something up; I'm pretty sure I did everything in the Contributing doc.


apps/playbox/playbox.yml
```diff
@@ -0,0 +1,14 @@
+name: Playbox
+description: 'An audio playback system for the live production industry.'
```
ckerr: [Don't start the description with A or An.](https://github.com/electron/apps/blob/master/contributing.md#submission-guidelines)

```suggestion
description: 'Audio playback system for the live production industry.'
```

cjdenio:
Whoops, I didn't see that. It should be fixed now.

PR #1062, opened 2019-05-15
chore: Update y2mp3 app's screenshots and description

moshfeu:
Some features are added so now the app is supporting not just mp3s


moshfeu:
Thanks!

PR #1064, opened 2019-05-19
feat: Add This My  PC  app   

supunlakmal:
ThisMyPC is a next-generation open source solution for remote pc. It's a full stack based project that starts as an experiment.


apps/thismypc/thismypc.yml
```diff
@@ -0,0 +1,12 @@
+name: ThisMyPC
+description: ' It is a full stack based project that starts as an experiment.Thismypc is software that you can use to browse your pc remotely from any browser
```
ckerr: Looks like CI is failing because the description is missing an endquote, so this suggestion adds that and shortens the description to a one-sentence explanation of what the software is used for.

Feel free to rephrase if this is _too_ short, but please keep it in the spirit of other app descriptions.

```suggestion
description: 'Browse your PC remotely from any web browser.'
```

supunlakmal: @ckerr  Ok thanks

PR #1065, opened 2019-05-19
Thismypc

supunlakmal:
ThisMyPC is a next-generation open source solution for remote pc. It is a full stack based project that starts as an experiment.


supunlakmal:
@jaco693 Yes you can test it. [http://thismypc.com](http://thismypc.com)

PR #1083, opened 2019-06-03
feat: Add Gihuba

Futureglobe:
Hi please add our app 'Gihuba'.
Because its a closed source app here is a screenshots proving that the app is made with electron:

image

Thanks


Futureglobe:
@ckerr hmm thank you for informing us.
Well the name means the following: GI (git) HU(hub) BA(backup).
If this name is a problem we will rename the app.
This would also require us to change all the backend stuff naming as of domain name and files we get over the network. But we will do it if the name is a problem.
Should we go on and change the name?

Thanks

PR #1085, opened 2019-06-04
feat: Add "Gibu"

Futureglobe:
Hi we renamed our app from 'Gihuba' to "Gibu" as requested from @ckerr.
Because its a closed source app here is a screenshots proving that the app is made with electron (we havent updated the screenshot because only the app name changed):

image

Thanks


ckerr:
Previous discussion: https://github.com/electron/apps/pull/1083

Futureglobe:
@ckerr CI is green now :)

PR #1087, opened 2019-06-04
build: prs are hard 😔

HashimotoYT:
Should fix releasing and should fix running builds from prs.


.github/main.workflow
```diff
@@ -17,7 +17,17 @@ workflow "Test" {
   resolves = ["Run tests"]
 }
 
+workflow "PR Test" {
+  on = "pull_request"
+  resolves = ["Run tests"]
+}
+
+action "Checkout pull request" {
+  uses = "siberianmh/git-checkout-pull-request-action@master"
```
HashimotoYT: I upstream this in sometime

BinaryMuse:
@HashimotoYT Thanks for tackling this.

I think the releases will still fail unless we replace `dotenv-safe` with `dotenv`, or use the `if (!process.env.GH_TOKEN)` check like we did in other projects.

BinaryMuse:
I also forgot to change `GITHUB_AUTH_TOKEN` to `GH_TOKEN` in `lib/github.js` 🙈

PR #1089, opened 2019-06-05
build: add automatic semantic title changer

HashimotoYT:
This PR adds automation script for changing standard title to semantic title for the pass all checks.

### Why?

> me: Good question. 😐

Mostly because when a user opens PR here not pass how minimum 1 check. It's one check it's semantic pull title, what's set user into a strange position, where last don't sure what's should do to pass to next level. (I think this description of the problem is true). This simplifies this by running its script when pr opened (or not 😂), also makes review process very small better, because the reviewer doesn't need manually update title. Robots make this automatically.

### But really, why?

I'm playing with GitHub APIs to making something strange and good, for robots and people, if it's not the very priority improvement, we can close this.

### And yep, examples 😊

See this pr list: https://github.com/HashimotoYT/apps/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc

### DIFF
--- a/.github/main.workflow
+++ b/.github/main.workflow
@@ -11,3 +11,14 @@ action "Update data and release" {
     "NPM_AUTH_TOKEN",
   ]
 }
+
+workflow "Update Pull Request title with Semantic" {
+  resolves = ["Update PRs"]
+  on = "pull_request"
+}
+
+action "Update PRs" {
+  uses = "actions/npm@master"
+  args = "run automatic-semantic"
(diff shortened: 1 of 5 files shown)


HashimotoYT:
For now, this action fails, because not installed `ts-node` when it's be ready to merge, I add `npm install` action.

ckerr:
I love that you're learning the APIs but this is PR is probably not something that should be merged.

PR #1099, opened 2019-06-16
feat: compress resized png icons

ckerr:
When we generate resized icons – the `$foo-32.png`, `$foo-64.png`, `$foo-128.png`, `$foo-256.png` files – pipe them through pngquant.

I've intentionally used a lossy compressor here because:
 * These are resized in a batch during `npm run build`, so CPU cost needs to be reasonable
 * These icons are already derived, so a lossy compression doesn't alter the original files
 * Choosing a lossy compressor gives a better speed/size result than lossless compression

Note that the original source icons are not part of this process. I've manually done lossless compression on them in #1077 and in this PR's first commit, but they are _not_ piped through lossy compressors.


zeke:
Nice! Any idea what the size differences are?

PR #1107, opened 2019-06-23
feat: add Pixel weather

Aveek-Saha:
This is a compact 8 bit style clear weather widget.

# Repo
https://github.com/Aveek-Saha/pixel-weather


apps/pixel-weather/pixel-weather.yml
```diff
@@ -0,0 +1,19 @@
+name: Pixel Weather
+description: 'Compact 8 bit style clear weather widget.'
+website: 'https://aveek-saha.github.io/'
```
ckerr: I'm of two minds about this. On the one hand, it's all to the good if open source devs can advertise themselves. On the other hand, this really isn't Pixel Weather's landing page -- it's your own professional page where Pixel Weather is an item down in the portfolio section.

If you were to make a dedicated PW landing page on your site, and that page _also_ linked back to your own page / portfolio, I'd be completely OK with that. But since the user has to hunt around to find Pixel Weather, I'm not going to accept this as-is.

Aveek-Saha:
Thanks for the feedback! I've changed the website link to a dedicated landing page

PR #1118, opened 2019-06-28
docs: Update Particl

LiteBit:


apps/particl/particl.yml
```diff
@@ -1,7 +1,7 @@
 name: Particl
-description: 'A Decentralized Privacy Platform with a suite of tools to enhance your online privacy'
+description: 'World's Most Private & Secure Crypto Marketplace'
```
ckerr: That apostrophe is breaking the yaml parser, which is why CI is failing for this PR.

```suggestion
description: 'World''s Most Private & Secure Crypto Marketplace'
```

PR #1122, opened 2019-06-30
feat: require ssl links for new apps

### LINKED ISSUE #1068: Require https for all images and website URLs
I saw some PRs land recently with `http://` URLs for the website.  It's 2019, ssl certificates are free and nowadays simple to set up (between lets encrypt, cloudflare and <insert hosting provider here> doing it for you) we should start forcing website URLs to be `https://`.

ckerr:
Fixes #1068.

Adds a test to confirm that all URLs in new apps' yml files are `https` or `sftp`.

As with some other tests, here we 'grandfather' in older apps that are already in the repo but which don't follow this requirement.

The new test's warnings look like this:

```
  153) human-submitted app data
       zoommy
         zoommy.yml
           should use ssl links:
     AssertionError: http://zoommyapp.com/: expected 'http:' to be one of [ 'https:', 'sftp:' ]
      at /home/charles/src/electron/apps/apps/test/human-data.js:103:65
      at Array.forEach (<anonymous>)
      at Context.<anonymous> (test/human-data.js:103:18)
      at processImmediate (internal/timers.js:439:21)
```

CC @MarshallOfSound


ckerr:
Instead of grandfathering, you mean? Several of the existing links don't work with https.

Interesting idea about winnowing down the list, though. We could fix anything that _can_ be shifted to https.

PR #1123, opened 2019-07-02
feat: Adding Infiniti Clips app

azure06:
[Here is the app URL (with its screenshots)](https://infiniticlips.com)

Please let me know if I've missed something!


azure06:
> LGTM. Thanks for the app!
Glad to hear that!

PR #1129, opened 2019-07-08
feat: Add r2e-todoworks app

tuantvk:


apps/r2e-todoworks/r2e-todoworks.yml
```diff
@@ -0,0 +1,18 @@
+name: r2e-todoworks
+description: 'Todo list r2e-todoworks built with Re-Mo-El.'
```
Sanjit1: Remove the full stop at the end and change ``` Re-Mo-El.``` to ```Re-Mo-El``` and the build will pass, and you will be able to add your app to the website

tuantvk: i fixed it, thanks so much !

PR #1145, opened 2019-07-20
feat: add test to ensure 'electron' is not a keyword

ckerr:
Minor PR to eliminate a minor annoyance: apps that include 'electron' as a keyword. Since *every* app in the repo is an Electron app, this is not a useful keyword.


apps/james/james.yml
```diff
@@ -5,8 +5,7 @@ repository: https://github.com/james-proxy/james
 category: 'Developer Tools'
 keywords:
     - proxy
-    - debugging-tool
-    - electron
+    - 'debugging tool'
```
codebytere: why was this put in quotes? should we be putting all other multi-part keywords in quotes as well?

ckerr: Force of habit. The real reason this line showed up in the diff was the hyphen removal.

I'm not a yaml expert but from [this backgrounder on whether or not to quote](http://blogs.perl.org/users/tinita/2018/03/strings-in-yaml---to-quote-or-not-to-quote.html) it looks like simple strings like the ones in our app metadata are safe to leave unquoted.

PR #1164, opened 2019-08-13
build: move to GitHub Actions CI

codebytere:
Reduces dependencies on TravisCI within Electron by moving `apps` to GitHub's own new CI/CI capable Actions infrastructure.

cc @ckerr


.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,19 @@
...
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@master
+    - name: Use Node.js 10.x
+      uses: actions/setup-node@v1
+      with:
+        version: 10.x
```
ckerr: Why does it have to be 10.x?

Asking to learn; not a PR blocker

MarshallOfSound: Don't think it has to be, just better to be specific, otherwise I think you just get latest LTS every time

codebytere: this actually does have to be 10.x, it uses a native module that won't compile on 12

PR #1168, opened 2019-08-15
build: convert schedule to actions v2

HashimotoYT:


HashimotoYT:
@MarshallOfSound not sure why this can block merge for this PR. The same thing doings Travis, it has Push, and PR jobs. Without adding "pull_request" event this can be blocked for merging PRs from users what's doesn't have CI job from "push" event, from our side we not able to do something for play around this.

codebytere:
I would say we can just leave it at Push for now.

HashimotoYT:
@codebytere I want to also see CI on `master` branch to verifying green status. I'm not see something bad with having two CI statuses, we do not open very more internal PRs.

codebytere:
@HashimotoYT in that event, i would say create two workflows, one for push and one for pull_request, and scope the former to `master` only?

HashimotoYT:
@codebytere 3c8d6bf9ba58b05a6716d65f4aff765ca69290a9 makes this pr happy?

codebytere:
@HashimotoYT this is syntactically invalid - `schedules` needs to be `schedule`

HashimotoYT:
@codebytere changed

PR #1172, opened 2019-08-23
feat: Account Surfer

dospolov:
Quickly switch between accounts and browsers with Account Surfer.

Decide what account or browser to use when opening the link.

Image of AS


dospolov:
@HashimotoYT sure.
image
image

PR #1177, opened 2019-08-30
app

nicfaust:
app


ckerr:
The old URL is still active, and the new URL appears to be a writing service that is not an Electron app. As such, I'm closing this PR.

If I've misunderstood, please open a new PR with more information.

PR #1209, opened 2019-10-04
feat: Add cashcash app

Winbee:
Cashcash is an accounting app whose goal is to assist the user with his personal budgeting/accounting. It is only available on MacOS for now.

URL: https://cashcashpro.com/


Winbee:
Sure, sorry I didn't see it before. Here is a gif showing the app launched in dev mode:

ezgif-2-b1fd339e674b

PR #1225, opened 2019-10-24
feat: add TestCafe Studio

kirovboris:
This PR will add [TestCafe Studio](https://testcafe-studio.devexpress.com) - Cross-Platform IDE for e2e testing.


kirovboris:
Thank you for the clarification!

image
Also we have `electron-updater` as a dependency in the EULA
https://www.devexpress.com/support/eulas/pdf/devexpress-eula-testcafe-studio.pdf

PR #1233, opened 2019-11-08
feat: Add PubvendorsTool app

EniroScott:


EniroScott:
image

Adding a screenshot. Really basic app but quite helpful for non-devs (marketing/ads teams etc) to generate pubvendors.json which is needed for new IAB compliant cookie-consent solutions that go into effect later this month.

PR #1240, opened 2019-11-15
feat: add debugtron

pd4d10:
[Debugtron](https://github.com/bytedance/debugtron) is a tool for in-production Electron based app debugging.

Screenshot


HashimotoYT:
@pd4d10 As temporary solution you can add your app at the end of this file https://github.com/electron/apps/blob/master/lib/grandfathered-descriptions.js with comment why he was added.

pd4d10:
@HashimotoYT  Thanks! Commit updated.

It is because the word "Electron" is necessary to describe the feature of Debugtron.

lib/grandfathered-descriptions.js
```diff
@@ -130,6 +130,7 @@ module.exports = [
   'data-store',
   'datazenit',
   'dbglass',
+  'debugtron',
```
HashimotoYT: Please add comment why he was added. This case is an exception than default action.

pd4d10: OK. Comment added

PR #1243, opened 2019-11-19
feat: add 'Dashy' app

Futureglobe:
Hi please add our app 'Dashy'.
Because its a closed source app here is a screenshot proving that the app is made with electron:

image

Thanks


Futureglobe:
@HashimotoYT only write access can resolve conflicts am i right? 
Do we need to do something?

Sanjit1:
> @HashimotoYT only write access can resolve conflicts am i right?
> Do we need to do something?

Fix the issues from your fork of electron/apps

PR #1244, opened 2019-11-21
feat: Update missive.yml

luismm88:
Changed the category to "productivity" and added 5 screenshots


missive.yml
```diff
@@ -0,0 +1,33 @@
+name: Missive
```
HashimotoYT: 🔥

luismm88:
Sorry about that, should be good now

PR #1295, opened 2020-01-18
feat: Add ExifCleaner

szTheory:
Website: https://exifcleaner.com/
Repository: https://github.com/sztheory/exifcleaner
Releases: https://github.com/szTheory/exifcleaner/releases

demo

ExifCleaner is a drag and drop GUI to remove image metadata. I think it's worth including because it's simple, it works, it's free, and I couldn't find anything else that did the same thing. I've been using it for more than a month and have fixed all the bugs that myself and other users (at least one from each platform of Windows, Linux, and Mac) have encountered during that time.


Toinane:
Hey @szTheory thank you for your application!
About the category, your application will better in Photo & Video category, what do you think?

szTheory:
@Toinane thanks good catch, I agree!

Toinane:
Its all good @szTheory thank you for your app! 👍

PR #1301, opened 2020-01-20
fix: Pragli logo and description

dsafreno:


dsafreno:
@HashimotoYT whoops, just pushed a fix

dsafreno:
Ah...  Just updated. How's it looking now?

PR #1306, opened 2020-01-28
feat: add Serene app

bigandy:
The application is built using Electron, React and Node. Here are some screenshots of the application with dev tools open.

Screenshot 2020-01-28 at 11 10 30

Screenshot 2020-01-28 at 11 09 03


ckerr:
```
  19649 passing (4s)
  1 failing

  1) human-submitted app data
       serene-app
         icon
           is between 256px x 256px and 1024px x 1024px:
     AssertionError: expected 1927 to be within 256..1024
      at Context.<anonymous> (test/human-data.js:214:42)
```

@bigandy if could you change the icon to be in the expected range, I think we'd be good to go

PR #1310, opened 2020-02-01
fix: Update wnr app

RoderickQiu:
Updated wnr's information.


RoderickQiu:
Thank you for working

PR #1318, opened 2020-02-09
feat: Remove Unicode Plus

tonton-pixel:
The Electron app `Unicode Plus` does not need to be listed any more, since it was actually renamed to `Unicopedia Plus` about one year ago, and disabled accordingly in the .yml file.

Users and search engines have been able to benefit from the GitHub automatic redirection during this rather long transition period, and I think it can be safely declared over from now on...

Thanks again.


tonton-pixel:
> Thank you for the PR but please follow the guidelines for the preferred way of removing an app:
> 
> https://github.com/electron/apps/blob/master/contributing.md#removing-or-disabling-apps

Thanks, but this is precisely what I already did one year ago when renaming the app from 'Unicode Plus' to 'Unicopedia Plus':

- Added a new `unicopedia-plus` entry,
- Added a `disabled: true` property to the `unicode-plus.yml` file.

The fact is that I had to rename the app to follow a request on behalf of the Unicode Consortium stating that the use of the name “Unicode Plus” was in violation of The Unicode Consortium’s trademark usage policy.

After one year, it is still possible to get results when searching for the old name, and also to directly access the old name page <https://www.electronjs.org/apps/unicode-plus>, which is somehow confusing, since downloads and release links do actually refer to the *renamed* app:

If it is indeed not possible to remove the app with the previous name, then I guess this PR can be rejected then, since it is already applying the recommended way of removing it by using the `disabled: true` property...

Thanks anyway,

-- Michel Mariani

PR #1324, opened 2020-02-25
feat: Add Norde Source app

gillisig:


gillisig:
Thank you @HashimotoYT, here is a screenshot to show that it is an electron app: 
image

PR #1334, opened 2020-03-03
chore: remove brave

codebytere:
We got an email this morning about Brave, and it indicated to me that the apps do not make it clear when they are not using Electron. I don't think there's much benefit to keeping them around when they're not Electron apps anymore, and it causes undue confusion. As such, i'd like to remove this.

cc @erickzhao


erickzhao:
Should we change our policy regarding app removal then? cc @electron/wg-ecosystem

malept:
I've added it to the WG meeting agenda.

PR #1379, opened 2020-04-10
docs: disables tidal

gitbreaker222:


gitbreaker222:
Thanks for checking this out. Yes seems you are right - sry for the mistake. I'm still in a phase of growing a feeling of what to expect seeing an apps page _(expected direct app links)_

PR #1459, opened 2020-07-03
chore: un-disable some apps

erickzhao:
I went through the apps that we disabled on first pass with dead links. These are repos that are still maintained and usually just had URL changes.


Toinane:
Looks good to me! 👍

PR #1277, opened 2020-01-03
chore: Merge and simplify GitHub Actions workflows

samuelmeuli:
The `pull_request.yml` and `push.yml` workflows run exactly the same scripts. This PR…

- Merges the two workflows
- Simplifies the `master` branch exclusion
- Uses a specific version of the `checkout` action (instead of using the `master` branch)
- Removes the `fetch-depth` paramter of the `checkout` action (default behavior since v2)

PR #1460, opened 2020-07-04
build: update actions config

HashimotoYT:
A good standard around our repos 🚀

Closes #1277
Closes #1279


.github/workflows/test.yml
```diff
@@ -0,0 +1,23 @@
...
+  pull_request:
+    branches:
+      - master
+  push:
+    branches:
+      - master
+
+jobs:
+  build:
+    name: Install dependencies
```
malept, line 13: The custom job name doesn't describe the job, only one of its steps. Copy/paste error?

```suggestion
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,23 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
```
malept, line 16:
```suggestion
      - uses: actions/checkout@v2
```

malept:
A GitHub admin needs to also update the protected branch settings, it seems.

PR #1474, opened 2020-07-17
Network Assembler added

kir486680:


apps/network-assembler/networkassembler.yml
```diff
@@ -0,0 +1,12 @@
+name: NetworkAssembler
```
HashimotoYT, line 1: Please rename this file to `network-assembler.yml`, and icon to `network-assembler-icon.png`

PR #1483, opened 2020-08-01
feat: add 'Super Mario'

dephraiim:


Toinane:
Hey @dephraiim Thank you for your application.
But I can't accept this application because it's use licensed assets from Nintendo and the name can be confusing with a Nintendo production. You need to change your application logo and the name before.
Also your applciation is too recent.

> If you just created something, wait at least 20 days before submitting.
Your repository created 7 days ago.

dephraiim:
Ok. Thanks.

PR #1499, opened 2020-08-15
chore: add utility script to find the archived repositories

vhashimotoo:
This PR adds the utility script to find the archived repositories and nothing more. Currently, I'm really not sure also what's we should do with them, delete or `disable`, since they are available but not more maintained.


Toinane:
Good PR! Maybe we can add a warning for these apps instead of disable it?

Issue #1548, opened 2020-10-01
Remove broken or deprecated apps


VerteDinde:
Electron Apps has collected a great deal of apps over the years. However, many apps are now broken or no longer working, and thus should be removed from our directories. As part of Hacktoberfest, we'd like to clean up some of those apps.

You can find a guide for removing/disabling apps here: https://github.com/electron/apps/blob/master/contributing.md#removing-or-disabling-apps

Criteria for a Broken App:
- Link to website no longer works
- Last release was over 4 years ago
- Using a version of Electron older than 3.x

Please limit PRs to removing a **maximum of 5 apps** per PR. This will allow others to also submit Hacktoberfest PRs, and keep our maintainers from becoming overwhelmed.

murtajaziad:
Hello, I would like to do this issue.
Can I start?

TDDR:
Hi, I would also like to contribute to this issue. I was just wondering if all 3 of the criteria need to be met in order to qualify as broken?

vhashimotoo:
@TDDR One criteria are enough to mark the app as delete candidate

MrOsta:
Hey, I found some apps that don't have an electron on their package.json , should i disable it?

Toinane:
> Hey, I found some apps that don't have an electron on their `package.json` , should i disable it?
Can you reference it here or in PR? If theses applications doesn't use Electron, they must be removed I my opinion

tussi003:
Ano

Toinane:
I've removed a lot of disabled application here #1740 and I'll make a new check soon, I suppose we can now close this 👍

PR #1552, opened 2020-10-01
fix: Disable 5 broken and deprecated apps.

a6macleod:

### DIFF
--- a/apps/bearychat/bearychat.yml
+++ b/apps/bearychat/bearychat.yml
@@ -4,4 +4,4 @@ website: 'https://bearychat.com'
 keywords:
   - chat
 category: 'Social Networking'
-disabled: true #Link to website broken
+disabled: true # Link to website broken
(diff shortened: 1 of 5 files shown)


ckerr (commented):
:+1: on disabling BearyChat, GF-Trader, and Zunburst -- all of those links are DOA for me.

PupaFM seems a little different to me -- the app is open source, still reachable, and it appears to still work. Even so, you could still make a case for disabling  it due to extreme age (ie it works, but it's running Electron 1.x).

I don't think we have any existing policy on where the cutoff point is for unmaintained apps. Whatever it is, I'd say 4 years is past the cutoff point :slightly_smiling_face: but CC @electron/wg-ecosystem in case anyone else has an opinion on this.

erickzhao (approved):
Thank you for this!

I went ahead and downloaded the PupaFM app and it doesn't seem to run for me on macOS. I'd be good with removing it from the directory!

vhashimotoo (approved):
Thanks for your contribution 💚.

Ref #1548

PR #1554, opened 2020-10-02
feat: Added entry for "Google Assistant Unofficial Desktop Client"

Melvin-Abraham:
**Repository:** https://github.com/Melvin-Abraham/Google-Assistant-Unofficial-Desktop-Client


vhashimotoo (changes requested):
As per https://github.com/electron/apps/blob/master/contributing.md#company-logos-and-names --

If `Google` has given you permission to use that name, please provide that information in the PR so that the app's name can be used as-is. Otherwise,

>

Melvin-Abraham:
Would it work if I change the application name in the yml file to **"Unofficial Desktop Client for Google Assistant"**?

ckerr:
> Would it work if I change the application name in the yml file to **"Unofficial Desktop Client for Google Assistant"**?

Yes, that would follow the guidelines. :+1:

Melvin-Abraham:
@vhashimotoo Updated application name in YML file...

Toinane:
Thank you for the name update! It looks good now 👍

PR #1606, opened 2020-12-11
feat: add screenshot

zonetti:


zonetti:
I wanted to add a screenshot because the README content was not being shown on the Electron Apps page, but it's working now so there is no need anymore.

PR #1607, opened 2020-12-11
feat: Added  Loop Team to App Directory

rolfrando:
Added Loop Team to the app directory.

Loop Team is a Electron app that supports Mac, Windows and Linux
See image for documentation that it is an electron app, or you may download app to verify at https://loopteam.co

101840785-9e2dc700-3af9-11eb-8351-e53d78d6c3f4


Toinane:
Hey @rolfrando thank you for your application! It looks good to me 👍

PR #1632, opened 2021-01-12
feat: YAGR now support English language.

gnu4cn:
Now the project YAGR-Yet Another Gitbook Reader support English lanugage throuth `@ngx-translate/core` and `@ngx-translate/http-loader`, other languages support also can be done, due to the flexibility of `@ngx-translate`.

And this commit fixes some description errors.

YAGR screenshot


Toinane:
perfect! 👍

PR #1646, opened 2021-01-17
feat: Add Erin app

losbiw:
[Erin](https://github.com/losbiw/erin) enables its users to change their wallpaper automatically
image
image


Toinane:
Hey @losbiw thank you for your app! It looks good to me! And sorry for the late response

PR #1648, opened 2021-01-18
feat: Add RBXUtil

JuiciiYT:
**RBXUtil** is an app which gets the user information of Roblox users, games and the catalog. The website can be found at [https://robloxutil.glitch.me](https://robloxutil.glitch.me). Here is a screenshot of it in use:
App Screenshot Here is our app logo:
App Logo


D1WK:
Hehe nice

JuiciiYT:
> Hehe nice

Thanks!

ckerr (commented):
Hey, just for clarification -- is this app endorsed by Roblox Corp?

If not, you may want to consider modifying the app's name due to possible trademark issues. The [electron/apps contribution doc](https://github.com/electron/apps/blob/master/contributing.md#company-logos-and-names) has some information about this:

>
>
> While some existing apps in the collection predate this rule and have been grandfathered in, we will not accept any apps that do not follow this rule going forward.

JuiciiYT:
No, its not but we have to have that name because that’s what it does so its easy for people to understand. Although we did ask Roblox but they didn’t reply.

JuiciiYT:
okay ill try

Toinane:
Hey @JuiciiYT Thanks for the update, but you need to update the folder name and your image's name too. Sorry for the late response

Toinane:
Hey @JuiciiYT any news on your PR?

Toinane:
As we don't have any news, I close this PR. Feel free to reopen it or make a new one

PR #1649, opened 2021-01-18
feat: Update cashcash yml

Winbee:
As the repo is now open-sourced, I've updated the meta data accordingly


ckerr (approved):
Thanks for the update!

PR #1652, opened 2021-01-22
fix: update Signal data

EvanHahn-Signal:
This makes three changes to Signal Desktop's listing:

1. Update to a newer icon
2. Remove `brew cask` listing, as it is not officially supported
3. Update to reflect our new license, AGPL-3.0-only


ckerr (approved):
Thanks for the update!

PR #1659, opened 2021-01-25
feat: Added Fireman app

yungookim:
Fireman is a Firestore and Firebase Auth desktop client that makes it easy to query, view, and edit Firestore documents with a traditional SQL-like GUI approach. It also comes with other features such as bulk CSV import, Firebase User listing, and data visualization (coming soon). Although it may appear similar to [Firebase Admin](https://www.electronjs.org/apps/firebase-admin) Fireman aims to have an approach as a data management client rather than a dev tool. 

Please let me know if any information is lacking. 

Thank you.

Built with Electronjs proof:
Developer Console is enabled with ipcRenderer shown in the console. https://getfireman.nyc3.digitaloceanspaces.com/screenshot.png

App screenshots
intro_page
main


ckerr:
Hi @yungookim, thanks for the app and for the information! LGTM.

ckerr (approved):

PR #1668, opened 2021-02-06
docs: update runjs description and url

lukehaas:
RunJS has moved to a new URL.


ckerr (approved):
Thanks for the update!

PR #1674, opened 2021-02-08
feat: Add Brisqi app

ashgrover:
This PR adds Brisqi app. Brisqi is built on top of Electron with React.

Brisqi is an offline-first personal Kanban app. It's available for Windows, macOS and Ubuntu (AppImage & Deb).
You can get Brisqi here -  https://brisqi.com

Here's the proof that it's written using Electron: 

image


ckerr (approved):
Thanks for the app!

Hey, this looks great! I know it's probably not a top-priority item, but are there any plans for a Linux build?

ashgrover:
@ckerr Thanks :) and yes it is available for Linux as I mentioned. It's available in AppImage and Deb format.

PR #1696, opened 2021-03-01
feat: add AfterLife app

AderitoNeto:
AfterLife allows users who have formatted their computer to install basic programs such as their favorite browser and their favorite game launcher.

There are only 3 categories here but there are more in the application.


Toinane:
Hey @AderitoNeto thank you for your app! Looks good to me! Just an idea, you should create your own logo, it would more cool than use an emoji 👍

AderitoNeto:
Thanks @Toinane for your Suggestion. I am working on that with many apps coming for the user.

PR #1706, opened 2021-03-05
feat: Add Nota

astoilkov:
[Nota](https://nota.md) is a writing app designed for local Markdown files. It's aimed at developers with IDE-like features and users that are building their own personal knowledge bases with features like wiki-links and backlinks.

We don't have a blog post describing our technology stack(we should) but I will try to provide proof it's built with Electron:
- if you download the app from https://github.com/notaapp/releases/releases/tag/0.23.1 you will see the size and its contents
- if you don't want to do it here is the contents of the Nota folder: 
image


Toinane:
Hey @astoilkov thank you for your application! Looks good to me 👍

PR #1707, opened 2021-03-06
feat: A password manager like app

HiroshiFuu:
My first Electron app. Replication of my C#.NET application that I used it for years to do fast pasting of passwords and tokens for different sites. This is an app that is missing from the electron apps.
https://github.com/HiroshiFuu/onecopy-electron


Toinane:
Hey @HiroshiFuu thank you for your application! Looks good to me 👍

PR #1721, opened 2021-03-25
Update nxshell.yml

Mizan-Al-Sabbih:


Toinane:
Hello, thank you for your pr! Do you have the permission from maintainer to change this information? Also, add a homepage link to electronjs.org isn't useful because when you click on this link, you are already on electronjs.org

PR #1727, opened 2021-03-27
feat: add tough cookies

LifeLongThinker:
Tough Cookies is a modern flash cards study and management app with Markdown and WYSIWYG support.
https://www.toughcookies.net

desk
manager
dashboard


ckerr (approved):
Hey, this looks good! Thanks for the app!

PR #1737, opened 2021-04-05
feat: Add LightyearVPN

lightyearvpn:


Toinane:
Hey @lightyearvpn thank you for your app! Since your application is a closed source, can you provide a proof that you use Electron? It can be a screenshot of package.json or devtools opened. Thanks!
Also, I've see your application is in English and Chinese, so you can add in your description yml "locale" property

lightyearvpn:
Hello, sure. We just updated our yml file and added locales property. Also, we just emailed you the proof. Thanks for your time.

Amanda888 (approved):

Toinane:
Hey @lightyearvpn thank you for the proof and update! It now all good to me 👍

Issue #1444, opened 2020-06-11
Delete disabled apps


erickzhao:
We should update the tooling in #1413 and do another pass to delete old, unmaintained application submissions rather than disable them.

This will help reduce the total package weight.

PR #1740, opened 2021-04-07
chore: delete all disabled and old applications

Toinane:
I've deleted all disabled apps as #1444 requested.
I checked that everything was good and that these applications are either too old or their link does not work anymore 👍


ckerr (approved):
LGTM. I sampled about a dozen of these removed entries and the removal all looks correct.

Thank you for doing this!

Toinane:
> LGTM. I sampled about a dozen of these removed entries and the removal all looks correct.
> 
> Thank you for doing this!

Yeah, I think it's time to make some cleaning here 😄

erickzhao (approved):
Thank you for this! 👏

PR #1750, opened 2021-04-09
chore: disabled old electron version app

Toinane:
I've used the new script to detect old electron version in package.json for apps with repository.
And here the result, I've checked manually if all was good!

I've added the submission guideline "use at least Electron v4.0.0"


erickzhao (approved):
The v4 cutoff is a bit arbitrary, but I think it's good to at least cull the apps a bit.

Wondering what everyone else thinks @electron/wg-ecosystem

Kilian:
Noooo one of my apps is in there. :sob: but this is a good idea! Do we need to make this a "rolling" requirement somehow, e.g. when 13 releases, the minimum becomes 5 etc?

Toinane:
> 
> 
> I think an occasional culling should work — no immediate need to update for each released version. If we would go for a rolling cull, we could have it only display apps with supported versions of Electron (last 3 majors) if we have enough of an ecosystem for that.

I'm not sure there would be many applications left if we only display those that are in the last 3 major versions 😅

PR #1753, opened 2021-04-10
feat: Added Socially

JackHumphries9:


JackHumphries9:
This one failed the build as I think it was because of some differences when I tried using the Wizard. For the new one, I just forked the repo and added the files on Github.

PR #1774, opened 2021-04-27
fix: name for google-chat-electron

ankurk91:
Hi,

These are the suggested changes

* There was a quote left in name of the app
* Removed "Electron" from app display name


Toinane:
Hey @ankurk91 thank you for the update! 👍

PR #1816, opened 2021-06-11
feat: Remove Snap option from Falix Software

KorbsStudio:
Due to complications with Snap errors, I've chose to stop updating the Snap version.


Toinane:
Hey @KorbsStudio thank you for the update! 👍

PR #1828, opened 2021-07-02
feat: Renamed Terminus to Tabby

Eugeny:
I've recently renamed it due to name clashing with another app: https://github.com/Eugeny/tabby/discussions/4087


ckerr (approved):
Thanks for the update!

PR #1846, opened 2021-07-25
fix: My Folder online

idoio:
I added the app repository & tags + fixed the broken image


Toinane:
Hey @idoio thanks for the update! 👍

PR #1860, opened 2021-08-18
chore: Fix's the broken image on SnailFM.

snaildos:
### Description
The app broke the CI, Very sorry for that, this fix should resolve it.


snaildos:
For some strange reason it edited other app meta files when running the check. I think it will be okay though?

PR #1932, opened 2021-12-09
fix: dbgate app update

janproch:
update of description and screenshot


JaydenDev:
@janproch 
Reminder:
You need a ```fix: ``` or ```feat: ``` prefix to pass the Semantic PR check.

PR #1984, opened 2022-03-18
chore: update repo maintainer

Toinane:
### Description

Update ownership of electron/apps to me.
I see a lot of ping of users to Zeke and it shouldn't be the case here.


erickzhao (approved):
Thanks!

PR #2026, opened 2022-06-13
feat: add bot-message-for-discord

Aryxstocratic:
### Checklist

- [x] I have read and completed the checklist of [Submission Guide](https://github.com/electron/apps/blob/master/contributing.md#submission-guidelines)
- [x] My application is open-source and has been created for at least 20 days
- [ ] My application is closed-source and here is a proof it uses Electron:

### Description

With discordbot you can send messages on discord using a bot, which can maybe be used for server purposes or just trolling. The installation and use guid can be found at the [source repository](https://github.com/Aryxstocratic/discordbot/blob/master/README.md)


ckerr:
https://github.com/electron/apps/blob/master/contributing.md#company-logos-and-names

Please do not directly use another company's name or product without permission. It's generally better to refer to it in a dependent clause; for example, after "compatible with", "on", or "for."

For example, while we would not accept a third-party app named "GitHub Notifications", we would consider "Yourname Notifications for GitHub".

While some existing apps in the collection predate this rule and have been grandfathered in, we will not accept any apps that do not follow this rule going forward.

Toinane:
Thank you for the change! Looks good now :)

Issue #2037, opened 2022-08-02
Strimio seems to not exist anymore


ropoko:
Both the site and the desktop application does not work anymore, I think it should be removed as the maintainer does not shared the source code and the project does not exist anymore.

erickzhao:
Thanks for flagging this @ropoko! It seems like they've posted an announcement about this downtime on Twitter, so I'm hesitant to remove the app altogether at the moment. I think temporarily disabling the app would be a good way forward.

https://twitter.com/strimioinc/status/1542200728437116928

ropoko:
I didn't see this, thank you

spiral2k:
Hey all, Strimio is back again, I've [opened a PR](https://github.com/electron/apps/pull/2073) to update the information.

@erickzhao thank you for not removing Strimio

PR #2038, opened 2022-08-02
chore: disable strimio

erickzhao:
ref https://github.com/electron/apps/issues/2037


erickzhao:
@strimio feel free to revert this commit when your app is live again!

